PR #5, opened 2015-09-11
Add `SourceTree` to developers

williamboman:


README.md
```diff
@@ -37,6 +37,7 @@
 - [Paw](https://luckymarmot.com/paw) - The ultimate REST client for Mac.
 - [PSequel](http://www.psequel.com/) - A PostgreSQL GUI tool.
 - [Sequel Pro](http://www.sequelpro.com/) - A MySQL database manager.
+- [SourceTree](https://www.sourcetreeapp.com/) - A free Git & Mercurial client for Mac & Windows.
```
iCHAIT: I think it would be better if the description is `A free Git & Mercurial client.` 
:)

williamboman: Yeah I thought so first too, but it might be worth mentioning it's cross-platform (well, Mac & Windows) for users who wants to use the same tools on both platforms?

iCHAIT:
Umm, in that case many of the applications on the list are cross platform, though they don't have it mentioned in their respective description so I think it would be better to stick to the structure.

Also for users who would be scrolling this list, I believe they would be searching for OS X stuff in particular.

Does that makes sense?

williamboman:
Sure, :+1:. Done.

iCHAIT:
:+1:

PR #13, opened 2015-09-11
Add `Amphetamine` to Utilities

williamboman:
This app is so much better than Caffeine, I sorta want to remove Caffeine from the list tbh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,8 @@
 ### Utilities
 
 - [1Password](https://itunes.apple.com/in/app/1password-password-manager/id443987910?mt=12) - Password Manager and Secure Wallet.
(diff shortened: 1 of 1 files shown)


iCHAIT:
@williamboman  Nice, we should have this on the list, but I am a bit reluctant in removing Caffeine from the list as it is really very popular among mac users, but on the same time it would also be redundant to have a couple of similar applications that serve exactly the same purpose in the list.

What do you think?

williamboman:
Since it is an "awesome" list, I definitely think Caffeine should be replaced by Amphetamine.

iCHAIT:
Hmm, fair enough, so you can remove Caffeine from this and after fixing the conflicts this should be ready to go :)

iCHAIT:
@williamboman So can you remove caffeine and fix conflicts?

williamboman:
@iCHAIT Done

PR #14, opened 2015-09-11
Remove locale and ambiguous params from iTunes URLs

williamboman:
This is probably a bit unnecessary and over the top.


iCHAIT:
:+1:

PR #17, opened 2015-09-13
InsomniaX

alexeyshockov:
Another great OS X app for preventing Mac from sleep. Like Caffeine and Amphetamine (from the adjacent PR).

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@
 - [DaisyDisk](http://www.daisydiskapp.com/) - Analyze disk usage and free up disk space.
 - [Finicky](https://johnste.github.io/finicky/) - App that allows you to set rules that decide which browser is opened for every link. ![Open-Source Software][OSS Icon]
 - [Helium](http://heliumfloats.com/) - A floating browser window that allows you to watch media while you work. ![Open-Source Software][OSS Icon]
+[InsomniaX](http://semaja2.net/projects/insomniaxinfo/) - Prevents OS X from automatically going to sleep (on lid closing or idle timeout).
 - [iStat Menus](https://bjango.com/mac/istatmenus/) - An advanced system monitor for your menubar.
 - [Kawa](https://github.com/noraesae/kawa) - A better input source switcher with shortcuts. ![Open-Source Software][OSS Icon]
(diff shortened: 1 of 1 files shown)


iCHAIT:
Hi, what are your thoughts on Amphetamine?

I think Caffeine should definitely be removed from the list as both Amphetamine and InsomniaX are much better.

Now between Amphetamine and InsomniaX,  I think it would not be worthwhile having two very similar application that serve the same purpose, so only one of them should be making to the list!

@williamboman  What are your thoughts on Insomniax?

williamboman:
I haven't used InsomniaX, but it looks pretty cool. I think I prefer Amphetamine though, just out of personal preference.

I noticed that InsomniaX is available through `brew cask`, so that's definitely a + for me.

iCHAIT:
@williamboman Okay, BTW I hope you'll agree with the idea of having one of these two on the list for the reason I have stated above?

williamboman:
I think it's worth including both InsomniaX and Amphetamine and explain what they do. I think InsomniaX does both "keep awake" and "stay awake if lid is closed, but go to sleep as normal", while Amphetamine only does "keep awake".

README.md
```diff
@@ -98,6 +98,7 @@
 - [Caffeine](https://itunes.apple.com/app/caffeine/id411246225) - Prevents OS X from automatically going to sleep.
 - [Finicky](https://johnste.github.io/finicky/) - App that allows you to set rules that decide which browser is opened for every link. ![Open-Source Software](media/oss.png)
 - [Helium](http://heliumfloats.com/) - A floating browser window that allows you to watch media while you work. ![Open-Source Software](media/oss.png)
+- [InsomniaX](http://semaja2.net/projects/insomniaxinfo/) - Prevents OS X from automatically going to sleep (on lib closing or idle timeout).
```
williamboman: lib -> lid

iCHAIT: :+1: 
Lets see what @alexeyshockov  thinks of this.

alexeyshockov: Thanks, fixed.

alexeyshockov:
Closed PR by mistake, sorry. Yes, I think having only InsomniaX in the list is more useful. As @williamboman said, this app solves both use cases and is available through Homebrew.

PR #18, opened 2015-09-13
Add Tower to `Developers`

sergiovilar:


iCHAIT:
@sergiovilar Don't you think gitup is better than tower?

iCHAIT:
Moreover, GitUp is open source and also Mac exclusive!

sergiovilar:
@iCHAIT I think that they are different things. Did you tried Tower? It's amazing :grinning:

Tower is also mac exclusive and is more like a "better" sourcetree, I think that is an alternative for who wants a paid app.

iCHAIT:
Oh, actually I didn't get the opportunity to use tower, though from the outset it looked like it was similar to GitUp.

Nevertheless, I think you may want to remove `mac` from the description and this should be ready to go :)

sergiovilar:
@iCHAIT done!

iCHAIT:
Thanks :)

PR #20, opened 2015-09-13
Add iStat Menus to `Utilities`

sergiovilar:


iCHAIT:
Same here, it should just be `An advanced system monitor for your menubar.`

sergiovilar:
@iCHAIT updated!

PR #21, opened 2015-09-13
Added DaisyDisk to `Utilities`

sergiovilar:


iCHAIT:
I think the description should be just - `Analyze disk usage and free up disk space.`, no need for mentioning `mac` :)

sergiovilar:
@iCHAIT makes sense, updated!

PR #23, opened 2015-09-14
added keka (un)archiver

vitorgalvao:


iCHAIT:
Thanks, was unaware of this :)

PR #24, opened 2015-09-14
TheUnarchiver: fixes

vitorgalvao:
As it were, it should’ve read either:

+ An app that unarchive**s** many different kinds of archive files.
+ An app **to** unarchive many different kinds of archive files.

Decided to follow neither and shorten the sentence. No need to say it’s an app (that’s what the repo is form), and the others do not follow that convention, anyway.

Also adds the OSS badge.


iCHAIT:
Hmm, makes sense :+1: 

BTW, I didn't know it was opensource, can you link to the source code?

vitorgalvao:
[On its homepage](http://wakaba.c3.cx/s/apps/unarchiver.html) (shouldn’t this repo be linking there, instead of the Mac App Store?) there is a link to download the source (box on the right side).

image

iCHAIT:
Thanks!

Yeah, I think linking the app to the homepage makes sense, can you change that?

vitorgalvao:
Done.

iCHAIT:
:+1:

PR #26, opened 2015-09-14
Add Adium IM client

nclements:
Adium IM client - slow (recently) on development but still going.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@
 ### Chat Clients
 
 
+- [Adium](https://adium.im/) - Multi-protocol IM client. ![Open-Source Software](media/oss.png)
 - [ChitChat](https://github.com/stonesam92/ChitChat) - A native Mac app wrapper for WhatsApp Web. ![Open-Source Software](media/oss.png)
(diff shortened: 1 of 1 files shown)


iCHAIT:
Hi, actually I was already aware of Adium (I use it regularly myself for IRC) but I didn't add it earlier itself because I feel Limechat and Textual are much better than Adium, also I had a chat with people on #osx and opened a thread on reddit for the same, and almost all of them preferred Textual or Limechat over Adium, so I didn't add it to the list, as this is an awesome list so only awesome items should go here.

What do you think?

nclements:
For the most part I'd agree, I prefer the interfaces of both. But Limechat and Textual are limited to IRC (or were when I last looked) while Adium supports Jabber, Yahoo!, Twitter, Facebook, etc.. So Limechat and Textual are great as IRC clients but less so for the rest (majority now?) of chat traffic.

But on the other hand Adium is now getting old and isn't as quick on the development. So closing the request.

iCHAIT:
Sums it up pretty well, thanks anyways :)

PR #33, opened 2015-09-15
Adding PhpStorm, PyCharm and IDEA

nahakiole:


iCHAIT:
I don't think this fits well into awesome os x, we don't want to end up having IDE's for each and every programming language.

PR #41, opened 2015-09-15
Add Divvy Window Manager

sunpech:


README.md
```diff
@@ -75,6 +75,7 @@
 - [Alfred](http://www.alfredapp.com/) - Boosts your efficiency and productivity.
 - [ClipMenu](http://www.clipmenu.com/) - ClipBoard History Manager. ![Open-Source Software][OSS Icon]
 - [CloudClip](https://itunes.apple.com/app/cloudclip/id563356503) - Sync your clipboard between your Mac and your iOS devices.
+- [Divvy Window Manager](https://itunes.apple.com/us/app/divvy-window-manager/id413857545?mt=12) - Window management for tiling your windows.
```
iCHAIT: I think you can link to - `https://itunes.apple.com/us/app/divvy-window-manager/id413857545` instead, shortens the link a bit.

sunpech:
@iCHAIT Alright, shortened it.

iCHAIT:
Correct me if I am wrong but isn't Divvy and zoom 2 (from your adjacent PR) similar in functionality?

If yes, them I would recommend having only one of them on the list.

kthibodeaux:
Why only put one, out of curiosity? All other 'awesome' lists (that I'm aware of) include all alternatives submitted.

iCHAIT:
Well, if two applications are very similar in nature and features then it just creates redundancy in the list and there is no point in adding both of them, having said that if their are significant differences in the features offered by them then of course we should include both of them. In the case here I am not sure but I think that both the applications are more or less similar.

@kthibodeaux Does that makes sense?

kthibodeaux:
More or less; I'm unaware of the two applications mentioned but if functionality and interface are that similar then I think you're taking the right approach.  Thanks for clarifying!

iCHAIT:
@kthibodeaux As I said, I not very much sure about these two applications as I didn't get the chance to use them and as they are not open source so was not able to have a closer look.

Lets see what @sunpech has to say.

sunpech:
@iCHAIT They are similar in some respect, but I use both as they are used for different needs.

I use Divvy when I want a window that fits in a grid layout that aligns with others. Basically for tiling windows in a grid layout.

I use Zooom/2 when I want a custom size window that I can resize without having to grab any corner-- say for example I want to see how a responsive website responds to being resized at 1900px, 1500px, 1200px, 800px, etc width. I also lets me move the entire window too.

iCHAIT:
Okay, I see, will dig a little deeper today for these two, BTW, can you resolve conflicts on your other PR's so that I can merge?

sunpech:
@iCHAIT I resolved them as best I could.

PR #45, opened 2015-09-15
Add Postgres.app

sunpech:


README.md
```diff
@@ -38,6 +38,7 @@
 - [Gitup](https://github.com/git-up/GitUp) - A simple but powerful Git OS X app. ![Open-Source Software][OSS Icon]
 - [ImageOptim](https://imageoptim.com/) - Makes images take up less disk space and load faster. ![Open-Source Software][OSS Icon]
 - [Paw](https://luckymarmot.com/paw) - The ultimate REST client for Mac.
+- [Postgres.app](http://postgresapp.com/) - The easiest way to get started with PostgreSQL on the Mac.
```
iCHAIT: You should be adding the OSS symbol to this as it is opens source.

dufferzafar: > On second thoughts I don't think this is something really Awesome!

Any particular reasons @iCHAIT ?

iCHAIT:
As this is an awesome list so we should only include awesome things, we don't want to add trivial applications and make this list like any other on the web.

brunowego:
Awesome tool :+1:

alexeyshockov:
May be http://www.psequel.com will be better? PostgreSQL itself don't corresponds with OS X and may be installed by many ways (I think, many developers prefer Homebrew for this), but PSequel is special application for OS X.

iCHAIT:
And psequel is already their on the list.

sunpech:
I find the comparison to PSequel to be incorrect. PSequel is a GUI tool so you can connect and run queries against a postgres database.

Postgres.app is the actual server/db itself. I know developers may use homebrew to install pg, but this is a nice alternative for many that want to get it up running fast. It's well supported and lets users turn on/off the server from the toolbar icon rather than the command line.

PR #49, opened 2015-09-16
Added RescueTime to the Productivity section

sheend:


iCHAIT:
I don't this fits well into awesome os x, primarily because it is web app.

And I am planning to include this is in the list as well - 

https://github.com/gurgeh/selfspy

It is really nice and open source too.

sheend:
Got it!

PR #55, opened 2015-09-16
Add ShiftIt to Utilities

brunowego:
:+1:


brunowego:
@iCHAIT I get conflicts with this branch and try solve it but I not know if I do it right. Can you explain if it is right?

PR #79, opened 2015-09-18
Add Sizeup to Utilities

swhite24:
Provide link to [SizeUp](http://www.irradiatedsoftware.com/sizeup/) under "Utilities" category.


iCHAIT:
@swhite24 Sorry for the late response, and thanks for the suggestion.

We already have [Spectacle](https://www.spectacleapp.com/) on the list which is definitely one of the best application in this category, so I am skeptic about adding SizeUp. Is there anything special about SizeUp that makes it different from Spectacle?

swhite24:
Hey no worries.  From what I can tell SizeUp isn't providing any additional utility on top of Spectacle.  Looks like a great tool!

PR #81, opened 2015-09-19
Added Slack to Communication

oscarmorrison:


oscarmorrison:
@iCHAIT

iCHAIT:
I don't think this fits the list as it has nothing to do with the Operating System itself.

herrbischoff:
I'm with @iCHAIT. Cross-platform UI apps should generally not be included.

oscarmorrison:
Ok thanks for your consideration @iCHAIT @herrbischoff 
I was under the impression that cross platform was acceptable, as you have plenty in the list already (Source Tree, Text Expander etc)

herrbischoff:
What I mean is cross-platform UI apps. SourceTree, TextExpander and others may be available for Windows as well but are well integrated in the Mac OS look, feel and technology. Applications like Slack are little more than wrappers around a web application. While this is a useful approach for this particular kind of application, it does not exactly stand out as an Awesome application. ;)

iCHAIT:
@oscarmorrison I hope @herrbischoff has explained it pretty well. You may close it :)

oscarmorrison:
Yes thankyou. @iCHAIT @herrbischoff

PR #120, opened 2015-09-21
Add CyberDuck and its CLI, duck

mikeanthonywild:


herrbischoff:
In my opinion, Cyberduck is everything but an awesome application. It's tedious to work with and pales completely in comparison with Transmit, which is already part of this list. @iCHAIT, what do you think?

iCHAIT:
I agree, Transmit is way ahead when compared to Cuberduck, @mikeanthonywild  thanks for the suggestion but please close the PR as it can't be accepted.

PR #122, opened 2015-09-21
Add Intellij Community Edition

FoxxMD:


iCHAIT:
I don't think this fits into Awesome OS X list as it is not a specific OS X application and has nothing to do with the OS itself. Thanks anyway :)

FoxxMD:
> not a specific OS X application and has nothing to do with the OS itself.

Under that criteria neither are:

* Atom
* Brackets
* Light Table
* Sublime Text
* Visual Studio Code

herrbischoff:
> Atom
> Brackets
> Light Table
> Sublime Text
> Visual Studio Code

Are all exceptionally made general purpose text editors. Intellij Community Edition is extremely limited in that it doesn't even feature JavaScript support. If we were to include every specialized editor under the sun, this would turn into a developer list in no time.

> Perhaps a new category for IDEs that are cross platform?

Nope. I for one feel that IDE's have no place here. With the exception of VSC no other entry you listed is an IDE. Otherwise there would have to be entries for Eclipse, WebStorm and countless other UX abominations as well.

PR #126, opened 2015-09-22
Add Mycli to Developers

veapon:
A awesome MySQL command line client!


herrbischoff:
This is a command line utility and — while absolutely useful — should not be included in this list. Please close.

PR #130, opened 2015-09-22
Adding Max to Audio section

bashu:


bashu:
squashed

herrbischoff:
The latest stable release of this software is from 2009. Even if most functionality may be included and it runs, this appears to be an awful long time of neglect.

bashu:
fine, but it works for me...

fedot:
Max is a great piece of software to have on the list. The fact that the release is from 2009 doesn't make it awful - it makes it stable.

BTW, unstable build is from 2011 and the development takes place on https://github.com/sbooth/Max (contributions welcome)

dkhamsing:
@herrbischoff should we include this to the list?

herrbischoff:
I'm still not convinced that this is a truly awesome addition.

PR #119, opened 2015-09-21
Add Slack to Chat Clients

AnatoliyGatt:
Slack - A messaging app for teams.

PR #131, opened 2015-09-22
Add Telegram to Chat Clients

AnatoliyGatt:
Telegram - A messaging app with a focus on speed and security, it’s super fast, simple and free.


vastus:
+1

herrbischoff:
Please do not use App Store links. The correct link for this application would be `https://telegram.org`. Thanks.

AnatoliyGatt:
@herrbischoff In #119 I was asked to replace app's homepage link with an App Store link. I guess the reason was that the app is not unique to OS X. So I'll keep it this way.

herrbischoff:
Good point.

PR #139, opened 2015-09-23
Add-Typora

abnerlee:
Typora- Another minimal Markdown editor.


herrbischoff:
Very nice app! I just tried it and the idea of seamless Markdown editing is fantastic. If you need a tester, count me in.

iCHAIT:
:+1:

PR #147, opened 2015-09-26
Dockphone and HipChat added

rahulkapoor90:


iCHAIT:
DockPhone looks pretty cool but I am a bit skeptic about HipChat.
@herrbischoff  Your thoughts?

herrbischoff:
I'd say that when Slack is accepted, HipChat should be as well — and vice versa. As both are mainly services and not OS X specific, I'd recommend against both of them.

iCHAIT:
@rahulkapoor90  I am closing this as of now. Please open separate PR's for each of your suggestions.

PR #148, opened 2015-09-29
Add Paste to Productivity

stel:


iCHAIT:
Looks to me that this is similar to ClipMenu, isn't it?
If yes, then this definitely seems to be way better than it, I think we should remove ClipMenu and have this on the list instead?

stel:
Yes, Paste is also a clipboard manager. As a developer I can't judge which is better, so I just pointing to this app :)

PR #154, opened 2015-10-03
Add LightPaper to Editors

ashokgelal:


iCHAIT:
@ashokgelal Hey, thanks for the suggestion.

We already have a number of Markdown Editors in the list,  can you throw some light as to how LightPaper is different/better than apps like [caret](http://caret.io/), [macdown](http://macdown.uranusjr.com/) and [typora](http://www.typora.io/) that are already on the list?

PR #176, opened 2015-10-25
Add MacGesture

CodeFalling:
MacGesture is a global mouse gesture application for Mac OS X.Open source at https://github.com/CodeFalling/MacGesture


iCHAIT:
I think this would be better off in the [Productivity Section](https://github.com/iCHAIT/awesome-osx#productivity).
Also you should place it in alphabetical order.

PR #182, opened 2015-10-28
Forecast bar

sunpech:


iCHAIT:
Nice app, any free alternatives for the same?

sunpech:
I used to use Degrees Weather http://www.degreesapp.com

But Forecastbar is much nicer.

iCHAIT:
:+1:

PR #193, opened 2015-11-07
Fixed el capitan # tag so setup link points to correct area of README

shaun10:
Due to space in El Capitan OS needed to add an extra dash to CSS inline link to correct point to area of our README file. Inline link now functions correctly, pointing to correct area of page when clicked.


iCHAIT:
Thanks

PR #217, opened 2015-12-30
Add Caret (markdown editor)

EdenSG:
[Caret](http://caret.io/) is a very capable but lightweight markdown editor.


iCHAIT:
@EdenSG  Already having [Typora](http://www.typora.io/) and [MacDown](http://macdown.uranusjr.com/)  on the list do you think that there is a need for Caret? I mean MacDown is a pretty solid markdown editor for mac and is also open source, while we also have Typora which is a lightweight one.

What do you think?

EdenSG:
Good question - perhaps there should be a *Markdown editors* section :smile: 

The thing is, they all serve the same purpose but in different ways. Caret, for example, seems even more lightweight than Typora, but it's got features that neither Typora nor MacDown have (such as some auto-completion and contextual commands). In my opinion it would be worth adding, if only because it'll suit some people better than other alternatives might (but the same could be said about those alternatives).

iCHAIT:
Ok, Fair enough, though I am not interested in making a separate section for Markdown Editors yet. Please resolve conflicts and I'll merge :)

PR #178, opened 2015-10-27
Add Blender

seanoh1989:

PR #218, opened 2015-12-30
Add Blender

EdenSG:
[Blender](https://www.blender.org/) is an extremely powerful full 3D suite, including 3D modelling, animation, sculpting, texturing, rendering, node-based compositing, video editing, etc...


EdenSG:
Just noticed: clone of #178, but Blender is FOSS

PR #222, opened 2015-12-31
Add Window Management category

EdenSG:
A lot of window managers were scattered in Utilities and Productivity, so they now have their own category.


iCHAIT:
:+1:

PR #224, opened 2016-01-08
Add Postico (PostgreSQL client)

thechunk:
A good looking (paid with trial) PostgreSQL client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@
 - [Insomnia](http://insomnia.rest) - A modern REST client with an beautiful interface.
 - [Integrity](http://peacockmedia.software/mac/integrity/free.html) - Easily find your website's broken links.
 - [Paw](https://luckymarmot.com/paw) - The ultimate REST client.
+- [Postico](https://eggerapps.at/postico/) - A modern PostgreSQL client.
 - [PSequel](http://www.psequel.com/) - A PostgreSQL GUI tool.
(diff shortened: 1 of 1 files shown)


iCHAIT:
@herrbischoff  Thoughts on this?

herrbischoff:
Difficult to say. This is the category of applications that need to be tested to get an impression. I'm really starting to get annoyed by App Store only apps.

deizel:
Been using this for a while now and it does the job (admittedly I don't work with Postgres every day).

Not sure if "trial" is the word I would choose, more of a "lite" version since it doesn't run out:

> **What are the limitations of the free trial?**
>
> - At most 5 connection favorites
> - Only a single window per connection
> - Table filters are disabled
> - There is no time limit — use the trial as long as you want!

dkhamsing:
@deizel 

dkhamsing:
let's close this at the end of the month if there is no activity

vassyz:
This one should definitely be in. My reasons are:
* I use it quite often, it's the most user-friendly Postgres tool you can get on macOS. Not as feature-rich as [pgAdmin](https://www.pgadmin.org/), but far easier to use.
* The author can be trusted, he already has a tool in the awesome-macOS list: [Postgres.app](http://postgresapp.com/)
* [PSequel](http://www.psequel.com/), the only Postgres GUI app mentioned here, is not as good as Postico, although it's free
* Postico can be used for free if you don't need more than five connections and don't have to use two connections side by side. Filters can be applied manually by writing the queries.

dkhamsing:
@vassyz thanks!

dkhamsing:
@thechunk can u rebase?

PR #231, opened 2016-01-29
Added M to Email Utilities

lksnmnn:
M is an open source email app for Mac (and iOS) with integrated device-to-device encryption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@
 ### Email Utilities
 
 - [Airmail](http://airmailapp.com/) - Lightning fast email client designed for El Capitan, and a free Beta.
+- [M](https://mynigma.org) - Lightweight email client with award-winningly simple, secure encryption. [![Open-Source Software][OSS Icon]](https://github.com/Mynigma/M)
 - [MailMate](http://freron.com/) - Advanced IMAP email client, featuring extensive keyboard control and Markdown support.
 - [Nylas N1](https://www.nylas.com/n1) - An extensible, open source mail client with custom plugin support. [![Open-Source Software][OSS Icon]](https://github.com/nylas/N1)
 


herrbischoff:
:-1: Just tried it out. While the built-in encryption may be nice (seems to be PGP-based), usability is really low. I feel that even modern webmail is more comfortable to work with. Definitely not awesome in my book.

lksnmnn:
Well, we’re working on that. Our main focus is/was building the crypto, key exchange and device sync functions. Additionally to that, we had to develope an usable email client (a real client, not those cloud based clients with a nice REST-API is a huge project and takes a lot of work and effort.

We keep improving both M for Mac and iOS and are currently working on an Outlook plugin and an Android app (where we can build upon K-9 Mail). We hopefully will have more developers and maybe a (UX) designer soon :)

herrbischoff:
Then, I believe the inclusion to this list should wait until the application is indeed well thought out and well designed. Personally, I couldn't care less about anything "Outlook"-related but I guess that depends on your goals. UX is paramount for an awesome application.

Interesting to read that you appear to be part of the M team. Let me express my critique of the application. As it stands now, M is just another uninspired attempt at an email client. I believe its concept is doomed precisely because it relies on either using the same application on both ends or installing some sort of plugin to make it work, minus the intermediate complexities like passwords that make PGP relatively safe in the first place. This route will see minimal adoption. Let me be clear on this: no user cares about the underlying technology. Not one. They do not care if it is more secure but less convenient. People chat via Facebook, send WhatsApp messages by the million and generally don't care too much about who may read their messages. I promoted PGP encryption 15 years ago, with the same arguments (like a postcard, etc.) Turns out, at the end of the day, most people do not seem to mind to basically send postcards. They value convenience over security. They are only going to switch if the new application is far better than the current one. M is a far cry from that.

As should be obvious by my wordy reply, I'm quite passionate about security vs. convenience. I highly value your ambitious attempt but ultimately see it falling flat on its face.

Just to put my money where my mouth is, I'd be happy to discuss and work on what I feel would make email work in this context. Maybe you're interested in such an exchange. Contributing to an email client that actually works would be great. Seriously, because I'm using `mutt` for years since truly all existing email clients suck.

lksnmnn:
Yes, I am one of the founders and joined the team a while ago. 

You are right and I know the app lacks awesomeness as a mail client. I, personally, also use Apple Mail too (which isn’t great either) since M misses some features I like to use. But I believe, looking only at the UX of end-to-end encryption, our first version is already a huge step into making PGP-like security more usable for non-IT folks. I mean those users who think "IE = Google = Internet“.

We currently working this project full time and want to transform it into a sustaining business, thus the Outlook-plugin. 

Both communication parties need to use the same underlying technology (for the encryption to kick in) in an app or a plugin, but so is Skype, Whats-App and Telegram… . I wouldn’t pin point this to a high probability to fail (it is a hurdle, but once overcome one of the main growth drivers).

At least in Germany you see more and more people gaining a conscious towards their loss of privacy and security when communicating online. The huge media attention after the NSA revelations lead the multiple new projects and drove the user adoption of secure messengers like Threema and Telegram (~ 60 million users). I think that 10 years from now, we all wonder how we could possibly send so much personal data unencrypted through the internet.

Some people will and do care about the underlying technology. That’s why we build an entirely new protocol which doesn’t rely on SHA1 and only uses large key lengths (4096-bit). We also encrypt subject lines, as most PGP and S/MIME users probably think they do. As for convenience, we are also leading the way, since the user does not need to know what a „key“ actually is. With our device pairing protocol, users can seamlessly pair their devices and synchronize their private keys to get access to secure messages on their mobile devices.

So security wise a huge step - email client wise, not **yet**. We always love to get feedback from people who really care about this topic and getting input on how to "fix email“ would help us a lot. As soon as we have some time, we will clean up our GitHub repository to make contributing more simple and appealing. A fair warning, building an email client and working with IMAP is a pain in the ass ;).

If you do like to get in touch, drop me a line at lukas [at] mynigma [dot] org.

iCHAIT:
@lksnmnn  All the best for M, however I agree with @herrbischoff that we should wait for merging M into this list for now, once it gets a bit matured and awesome then we should definitely have this on the list.

I hope that is fine with you :)

lksnmnn:
Of course. I see your point :) 
Just a reason more to keep working!

iCHAIT:
:+1:

herrbischoff:
> If you do like to get in touch, drop me a line

Will do! I'm pretty skeptical about turning email security into a sustainable business though.

iCHAIT:
@lksnmnn  Do you mind closing this PR for now, and reopening it once you think it is ready for it?

PR #235, opened 2016-02-08
Added oh-my-zsh

highvoltag3:
Added oh-my-zsh, wondered if this should be under terminal instead but since there's already a ZSH entry here I figured it'd be consistent.


herrbischoff:
This has nothing to do with Mac OS X and is not even an application *per se*, as it it a script-based shell framework. Please close.

PR #241, opened 2016-02-09
Added Image2icon

highvoltag3:
Nice little utility to create icons for Mac. Great for quickly
exporting retina versions for development, but could also produce icons to customize your Mac.

PR #242, opened 2016-02-09
Added ImageOptim

highvoltag3:
ImageOptim is a free app that makes images take up less disk space and
load faster, without sacrificing quality.


highvoltag3:
@iCHAIT Might wanna approve #241 first to keep the alphabetical order.

iCHAIT:
@highvoltag3  ImageOptim is already on the list under Developer section.

highvoltag3:
Completely right my bad.

PR #245, opened 2016-02-09
Added CleanMyMac

highvoltag3:
Added clean my mac utility been using this for years since it’s first
version and I think it’s does what it states really well.


iCHAIT:
@herrbischoff Why a 👎 ? I actually used it a while back and found it pretty reasonable.

dkhamsing:

herrbischoff:
IMHO, all commercial "cleaning" and "optimization" tools are a rip-off because macOS runs most maintenance tasks itself. Besides, any person who is reasonably proficient with GitHub should be proficient enough with their OS to not feel the need for such snake oil. This isn't Windows country after all.

dkhamsing:
I think I agree with @herrbischoff 
Closing

PR #253, opened 2016-02-18
Added Infinit to Sharing Files section.

cadeuh:
Added Infinit to Sharing Files section.


dkhamsing:
Free if sharing with other Infinit users..

@cadeuh We have quite a few other file sharing apps right now on the list.. why should we add Infinit?

herrbischoff:
I'd close this. It's just another tool that does what everyone else is doing as well.

dkhamsing:
ok

PR #255, opened 2016-02-19
Update Neptunes: no longer open source

ReadmeCritic:
https://github.com/rurza/NepTunes is gone, there is now https://github.com/rurza/NepTunes-Issues


iCHAIT:
Fixed in another PR.

PR #272, opened 2016-03-14
Add Paparazzi

herrbischoff:


herrbischoff:
How about this?

iCHAIT:
Another application that I am somewhat skeptic about, I was wondering how is it different from [cloudapp](https://www.getcloudapp.com/) and [mac2imgur](https://github.com/mileswd/mac2imgur) both of which I use and are pretty awesome in all aspects. Also from the screenshots on its website it looks like the application does not support El Capitan I guess.

What do you think?

herrbischoff:
It's different in the fact that it creates a screenshot from a *website*, not your desktop, and has no upload functionality whatsoever. ;)

iCHAIT:
Okay, so I have 2 points to make here - 

1. With `cloudapp` and `mac2imgur` we can take screenshots of our desktop as well as any website (or any other thing for that matter).
2. Secondly, don't you think that not having an upload functionality is a drawback in itself?

herrbischoff:
I believe we are talking about different functionalities here.

Regarding 1, allow me to demonstrate:

how people build software github 20160330

Can any of the current apps create such a full-page screenshot of a website? Or something like that for the matter:

jquery to vanilla js 20160330

Regarding 2, I see no problem because it features the standard OS X sharing pane:

screen shot 2016-03-30 at 19 26 58

It just doesn't handle uploading itself. Therefore, you can use any service you want, provided a sharing extension for OS X exists. Every Awesome application in the Sharing section should have that, I believe. ;)

iCHAIT:
Ah sweet, got your point, the second example is right on the money. Thanks for explaining it in detail.
Merging :+1:

herrbischoff:
Glad that you like it!

iCHAIT:
Hey @herrbischoff ,  I actually used this tool today for the first time ever, and I must say it is a very neat application, works great and has a number of useful features.

Indeed and awesome addition to the list :tada:

Thanks 👍

herrbischoff:
@iCHAIT, good to hear that it fit the bill for your needs! :)

PR #275, opened 2016-03-27
Add Avocode to Graphics

AnatoliyGatt:
Avocode - The bridge between designers and developers. A way to share designs, export assets and collaborate with your team.


herrbischoff:
As this is a subscription service rather than a stand-alone application, I'm skeptical.

AnatoliyGatt:
@herrbischoff It's got a client app for Mac, which is well made. I believe that the developer's revenue model shouldn't be the reason why to accept or reject its listing in the repo.

iCHAIT:
Needs +1

PR #276, opened 2016-03-31
Add Franz

mence:
[Franz](http://meetfranz.com/) wraps a number of chat services into a single GUI application, which means less open tabs/applications!

It currently supports: 
- Slack
- WhatsApp
- WeChat
- HipChat
- Facebook Messenger
- Telegram
- Google Hangouts
- GroupMe
- Skype
- Grape

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@
 
 - [ChitChat](https://github.com/stonesam92/ChitChat) - A native Mac app wrapper for WhatsApp Web. [![Open-Source Software][OSS Icon]](https://github.com/stonesam92/ChitChat)
 - [Colloquy](http://colloquy.info/) - A GUI IRC Client. [![Open-Source Software][OSS Icon]](https://github.com/colloquy/colloquy)
+- [Franz](http://meetfranz.com/) - Combines chat & messaging services into a single application.
 - [Goofy](http://www.goofyapp.com/) - Unofficial Facebook Messenger client. [![Open-Source Software][OSS Icon]](https://github.com/danielbuechele/goofy)
 - [LimeChat](http://limechat.net/mac/) - Instant messaging application. [![Open-Source Software][OSS Icon]](https://github.com/psychs/limechat)
 - [Messenger for Desktop](http://messengerfordesktop.com/) - An app for Facebook messenger. [![Open-Source Software][OSS Icon]](https://github.com/Aluxian/Facebook-Messenger-Desktop)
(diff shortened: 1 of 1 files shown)


herrbischoff:
Another Electron wrapped web application…

mence:
@herrbischoff I think Franz uses it well - it wraps the web clients of the various services which means you still get a pretty full featured experience. The alternative I suppose would be writing a native implementation of each service. I find Franz pretty useful, but I understand the dislike of "webapps wrapped up" approach that most Electron apps take.

mence:
The broken build looks like pre-existing failures:
```
Issues :-(
> Links 
  1. 503 http://sourceforge.net/p/zsh/code/ci/master/tree/
  2. 503 http://sourceforge.net/p/burn-osx/code/HEAD/tree/
  3. http://burn-osx.sourceforge.net/Pages/English/home.html Connection reset by peer
  4. 503 http://sourceforge.net/projects/keka/files/
  5. 503 http://sourceforge.net/p/wineskin/code/ci/master/tree/
```

herrbischoff:
Good to see that you understand where I'm coming from. I feel this list should feature the best Mac OS X has to offer (just read that it may be re-branded to macOS, analog to iOS/watchOS/tvOS, in the future), not cross-platform implementations of the same HTML-based interface within a closed-off silo of itself. As always, @iCHAIT has the final word on this. ;)

What I would be most interested in right now: how does Franz solve the problem of connecting to Skype? I cannot seem to find any clear information on that on their website. As it does not appear to be open source, I'm confused about that. Usually this is done via proxying or by connecting to a locally running Skype instance. The former would be troubling, as it reflected on the recent trend of applications that look like free apps on the surface but in reality are services through will all your communication flows. Exhibit A: Nylas N1 email client. I am not especially keen on spreading my personal information any wider than it is already spread.

iCHAIT:
Well, frankly I am not sure about this one, @mence it is a great application indeed, but as @herrbischoff  rightly mentioned, at the end of the day it is a web application and as far as this list is concerned we are more inclined to include GUI applications rather then web applications as the latter ones are generally cross platform and are usually independent of the OS, so we are a bit skeptic to include such applications in this list. 

Having said that as Franz seems to me a great application, so I guess we should rethink our strategy, so I guess the best bet f or this would be to have thoughts of other people on this maybe? If this manages to get a couple of +1's I'll merge this. Does that sounds fair @mence @herrbischoff ?

herrbischoff:
@iCHAIT: Sounds good to me. Changing development landscapes should always be considered. A native application that bundles all kind of messengers would be far preferred through. I already use Slack exclusively via IRC gateway, because IRC is where I am. :)

Still, I would wait until the privacy issues are cleared up. I for one would't want to recommend an app that is shady on how it works and where your data flows.

iCHAIT:
Needs +1

PR #287, opened 2016-04-29
costless badge

briatte:
just the Creative Commons NonCommercial badge with the transversal bar
flipped horizontally — might be used to identify free (as in free beer)
software


iCHAIT:
Fixed by #294

PR #297, opened 2016-05-06
Remove MPlayerX

faviouz:
MPlayerX's official installer contains malware/adware. I would argue this isn't _awesome_. I suggest removing it from the list.

https://www.google.com/#q=mplayerx+malware


iCHAIT:
Thanks for reporting and opening the PR 👍

PR #298, opened 2016-05-08
Added cloud magic to email clients

Jarvvski:
Not sure why my editor removed the white space on background music. Still, should merge fine.

Note: Cloud magic does charge for their Profiler addon that looks up a senders linkedIn accounts etc if it exists. Full email functionality is free.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@
 
 - [Audacity](http://audacityteam.org/) - Free, open source, cross-platform software for recording and editing sounds. [![Open-Source Software][OSS Icon]](https://github.com/audacity/audacity)
 - [Audio Hijack](http://www.rogueamoeba.com/audiohijack/) - Record audio from any application like iTunes, Skype or Safari, or from hardware devices like microphones and mixers.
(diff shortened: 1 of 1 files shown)


dkhamsing:
$20 on the app store https://itunes.apple.com/us/app/cloudmagic-email/id1059655371
can anyone else comment on this app

Jarvvski:
Strange, did they update with a price? I certainly never paid for it however that might have been during the beta. I cannot remember fully.

Either way the fact that it stores your settings for you on your iCloud allows me to easily move over to new devices or manage all my emails from any apple device. (Having 7+ email accounts gets hard to keep adding them per device)

dkhamsing:
Are you sure your credentials are stored in iCloud? I thought they were stored with cloud magic
http://www.theverge.com/2016/1/20/10797070/cloudmagic-mac-email-app-review

If that is the case, this feature may or may not be awesome, security wise..

Jarvvski:
Hmm. It might be best to close this pull until I/we have looked into this more. It's been a while since I first installed it, and I still very much enjoy using it. However, to best describe it in terms of cost and underlying functionality I'd need to focus on it for once rather than just going throught the daily email grind

PR #300, opened 2016-05-14
NepTunes re-added.

rurza:
It was removed because I've made repository private. But app still exists and it's available to buy.


herrbischoff:
👎 because of the switcheroo from open source to paid.

rurza:
So what? Isn't awesome-osx about great apps? Paid, free, open source - doesn't matter.

herrbischoff:
Despite your snappy "so what" retort, I'm going to take some time out of my schedule to write up why I feel this way in a more thorough fashion.

I for one do not like projects which start as FOSS just to suddenly remove the source completely to become paid only. This is a big no-go for several reasons. At the very least the original source code should still be available since it was licensed as FOSS originally. Charging for a more polished version is quite alright (one has to make a living or earn some spare change at least), removing source code altogether is not – at least in my book. Either leave the latest code version before going closed source available, even if you never update it again, or never release the source in the first place. Both is fine. Take a look at the Textual IRC client for an ideal approach: current source is free and open, compiled version is 5 bucks in the App Store to support development. If your app is useful enough to people, they will pay to support further development. If you're trying to take away their alternatives and pressure them to update, they will not.

Since it's your work and your code, you are free to do whatever you want of course. You just need to be prepared how people react to you being – what in the FOSS community is quite uniformly considered to be – a jerk. Furthermore, if you ever accepted just a single pull request, no matter how small, it's not entirely your code anymore. The author of that pull request may want their work to remain free, open and available. It was submitted under the license that was current at the time. Technically and depending on the original license, you could even be in breach of said license which usually requires the code to be available at all times. Specifying a liberal free software license does not enable even its original author to suddenly revoke said license retroactively. Also, again depending on the type of license, commercial building and improving upon without publishing the modified source may be fine (in case of MIT, for example) or even explicitly forbidden (in case of a GNU license). In any case, removing the original source is a jerk move.

This is my opinion and I hope it clears up my thumbs down.

PR #305, opened 2016-06-09
Add NeoVim Dot App to the editors list

wazery:


herrbischoff:
👎 

While it's a commendable effort, it is in many ways worse than MacVim. Have you seen it's scroll performance — it's abysmal. So this is definitely not an awesome application. Yet.

PR #306, opened 2016-06-09
add StreamJockey to Others

nogizhopaboroda:


dkhamsing:
has anyone else tried this?

herrbischoff:
I haven't tried it because the concept appears to be too complex and really solves a non-existing "problem". To me, it looks like another of those NodeJS-based terminal applications that exists solely because it can.

PR #308, opened 2016-06-13
Add Canary email client

EdenSG:


herrbischoff:
This is beta software, not sure if this should be on the list. Looks like an almost 1:1 clone of Airmail. Also, please add a period to the end of the description.

dkhamsing:
Closing for now.. maybe we can reevaluate when it's out of beta

PR #311, opened 2016-06-24
Add CurrencyX to Utilities

TomLiu:


dkhamsing:
It looks beautiful and costs $2 
Can anyone else comment on this app?

iCHAIT:
I have not tried it out but it looks good.

dkhamsing:
@TomLiu 

herrbischoff:
Is there really a need for a dedicated currency conversion app when all you need to do is Google or DuckDuckGo something like `10 usd to eur`?

dkhamsing:
hm good point
let's close for now

TomLiu:
Sorry to reply so late. I have fixed the conflicts. 
Someone like do everything through the browser, someone don't I think.

herrbischoff:
@TomLiu: Sure, but what is the point of having such a (paid non the less) tool at all when you can achieve the same for free? Not everything needs to be an "app", many hobby developers forget that. You need to improve the current status quo with your application or keep it free as an object of learning. If it were free and/or provided *significant* improvements over the browser workflow, it could be considered "awesome". As is, it's just a GUI wrapper for a trivial internet request.

TomLiu:
@herrbischoff you can use gmail for free, why a lot of people pay a lot of money for a native mail client like Airmail, Sparrow?

dkhamsing:
regardless of price, at this point we don't consider your app to add that much value. sorry.

herrbischoff:
> why a lot of people pay a lot of money for a native mail client

Precisely because those mail clients *add significant value*. Your app does not.

PR #318, opened 2016-07-01
Add Dr. Cleaner to Utility

GingerLi:
Signed-off-by: Ginger <njupt88@163.com>


GingerLi:
Do I need to do something for the Check Failure? It seems not related to my change.

herrbischoff:
The `&ct=Github.osx` suffix on the URL just reeks of tracking, making me very suspect about the motivation behind this addition. It smells like a strategically placed addition by the developer (Trend Micro) or someone close to them. The application is free but with IAP and is in the rather useless category of "optimization apps". 👎 from me.

PR #321, opened 2016-07-05
Add Jump Desktop

haseebq:
Jump Desktop is a remote desktop (RDP and VNC client). It's available on the Mac App Store: http://itunes.apple.com/us/app/jump-desktop-remote-desktop/id524141863?ls=1&mt=12

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,7 @@
 - [Helium](http://heliumfloats.com/) - A floating browser window that allows you to watch media while you work. [![Open-Source Software][OSS Icon]](https://github.com/JadenGeller/Helium)
 - [InsomniaX](http://semaja2.net/projects/insomniaxinfo/) - Prevents OS X from automatically going to sleep (on lid closing or idle timeout).
 - [iStat Menus](https://bjango.com/mac/istatmenus/) - An advanced system monitor for your menubar.
+- [Jump Desktop](https://jumpdesktop.com) - A powerful RDP and VNC client with built-in support for SSH tunnelling and zero config setup.
 - [Kawa](https://github.com/noraesae/kawa) - A better input source switcher with shortcuts. [![Open-Source Software][OSS Icon]](https://github.com/noraesae/kawa)
 - [KeePassX](https://www.keepassx.org) - A free, open source, and light-weight password manager. [![Open-Source Software][OSS Icon]](https://github.com/keepassx/keepassx)
 - [KeepingYouAwake](https://github.com/newmarcel/KeepingYouAwake) - Menu bar utility that prevents Mac from going to sleep. [![Open-Source Software][OSS Icon]](https://github.com/newmarcel/KeepingYouAwake)


dkhamsing:
$29💰

dkhamsing:
@haseebq why is Jump Desktop a good app?

haseebq:
@dkhamsing I work on Jump Desktop, so I'm biased. That said - here are some of the reasons why I think it's awesome:

- It lets you connect to any computer regardless of the OS it's running: It natively speaks the RDP and VNC protocols which means that it has the ability to connect to almost any machine without the need to install software on the remote side (Windows over RDP, Macs over VNC/ARD, Linux over VNC). Further more is also supports RD Gateways (for Windows users).

- It has a built-in SSH client to secure VNC connections. It'll automatically log you into your SSH server and create the forwarding to your host.

- It's energy efficient: We use hardware accelerated rendering and actively reduce cpu/network usage when a connection is not in use.

- Customizable keyboard mappings: When connecting to Linux/Windows, you can use your standard Mac shortcut keys and Jump will automatically translate these for you. For example Cmd+C turns into Ctrl+C and Cmd+Q into Alt+F4.... it's all customizable. This is a big productivity booster for hard core Mac users who work on Windows/Linux machines because it frees them up from mental burden of switching keyboard shortcuts while they do their work.

- It supports multi-monitors both locally and remotely. This works especially well if you're using RDP to connect.

- An optional app called Jump Desktop Connect that you can install on the remote server. It makes setup super easy and lets you connect from anywhere (LAN or WAN) without the need to worry about IP addresses, dynamic/static dns or port forwarding. Our goal was to make it so easy to setup that non-technical users can set it up in 5 minutes. We use it internally for all our machines because it's so much easier to connect without worrying about DHCP/IP addresses.

- iCloud syncing: Supports syncing connection entries across your Macs and/or Jump Desktop for iOS. We've also publicly our server file format here: https://support.jumpdesktop.com/entries/38046030

- Fluid Remote Desktop: It's a new protocol that we're going to be releasing this summer and will be a free update for all users. We're aiming to make it much more efficient and capable than RDP or VNC. Encryption is built-in, it supports full motion video streaming, multimon support, has rich clipboard integration, audio streaming (planned) touch redirection (planned) etc. Watch a quick demo here: https://www.youtube.com/watch?v=QtdTCeB1eeg .

Let me know if you have any questions.

haseebq:
Also, not sure if it helps, but there is a trial version available here: https://jumpdesktop.com/downloads/jdmac

dkhamsing:
Thx 
Trial certainly helps

dkhamsing:
I don't really use these types of apps myself..
@herrbischoff what do you think

herrbischoff:
Since it's MAS-only there's no way to try it. I professionally used several different VNC clients in the past with [Screen](http://edovia.com/screens/) standing out. That one is Mac-only though. RDP support sounds nice to have but lacking a Windows server (I'm a Unix guy) and the MAS restriction would put this in the "I don't know, maybe wait for possible additional comments for a month" category.

haseebq:
It's actually available outside the MAS too as a trial. Here's the download
link: https://jumpdesktop.com/downloads/jdmac

Should I change he pull request to use the trial link instead of the MAS
link?

herrbischoff:
@haseebq: No, it's alright. We'd just like to hear from others if they think your application is indeed noteworthy. We're going to wait for a couple of weeks and see what comes up.

dkhamsing:
We feel this change did not receive enough endorsement
Sorry 😢

PR #323, opened 2016-07-12
Add Telephone

eofster:
Telephone is a SIP softphone that is compatible with virtually any SIP provider and allows you to make and receive calls.


herrbischoff:
Good application: simple, functional, gets out of your way. 👍

iCHAIT:
👍

PR #334, opened 2016-07-15
Add links to 2 CLI tools

dufferzafar:


herrbischoff:
I'm absolutely no fan of simple wrappers that amount to little more than a collection of functions. 👎

dkhamsing:
If you live in the terminal, these cli tools might be pretty useful

iCHAIT:
+1 from me too

herrbischoff:
@dkhamsing: I do live in the terminal but I dislike these helper function collections. Judging from the reactions here and all over GitHub, I seem to be in the minority though.

dkhamsing:
@herrbischoff you have very strong opinions, I like that

herrbischoff:
> you have very strong opinions, I like that

I have a rather Jobsian approach to the things I do (or not do):

> “People think focus means saying yes to the thing you've got to focus on. But that's not what it means at all. It means saying no to the hundred other good ideas that there are. You have to pick carefully. I'm actually as proud of the things we haven't done as the things I have done. Innovation is saying no to 1,000 things.”

PR #285, opened 2016-04-26
Add Batch Image Resizer

brunowego:

PR #340, opened 2016-07-19
Add Batch Image Resizer

brunowego:
Ref. #285


dkhamsing:
@brunowego is this free?

brunowego:
The site says nothing about license.

dkhamsing:
@iCHAIT 
@herrbischoff 
should we add the free icon?

iCHAIT:
I think its free, their site mentions -

> While this application is provided free of charge, you can support us by purchasing it on the Apple App Store for just $2. 

So I believe we can add freeware icon to it.

PR #349, opened 2016-07-24
Added Dropzone to the productivity section

aptonic:
Just found this repository and think it's really neat. Have added a paid productivity app that I'm the developer of. It's available both in and out of the Mac store and also has a 15 day trial.

Project URL: https://aptonic.com/

Why should this project be added: It's a fairly popular Mac App - see here: https://www.macupdate.com/app/mac/31794/dropzone

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,7 @@
 - [BetterTouchTool](https://www.boastr.net/) - Configure gestures for mouse and actions for keyboard shortcuts.
 - [ClipMenu](http://www.clipmenu.com/) - ClipBoard History Manager. ![Freeware][Freeware Icon]
 - [CloudClip](http://www.thinkbitz.com/cloudclip/) - Sync your clipboard between your Mac and your iOS devices. ![Freeware][Freeware Icon]
+- [Dropzone](https://aptonic.com/) - Create a popup grid of customizable actions that enhance productivity on your Mac.
 - [f.lux](https://justgetflux.com/) - Automatically adjust your computer screen to match lighting. ![Freeware][Freeware Icon]
 - [Fantastical 2](https://flexibits.com/fantastical) - Complete Calendar app replacement which uses natural language for creating events.
 - [Hazel](https://www.noodlesoft.com/hazel.php) - Create rules to automatically keep your files organized.


herrbischoff:
@aptonic: I purchased Dropzone a while ago, as it was heartily endorsed by Brett Terpstra aka Mac Productivity Nerd. However, it quickly became stale and I was struggling to find any use at all for me. The most interesting one was the file upload drop but since Transmit does it as well, I didn't use it more than a few times. Its scriptability was appealing to me as well, still I never found a use for it.

Maybe you could share interesting uses for Dropzone I may certainly have overlooked.

aptonic:
@herrbischoff Have a look at the third party action list at https://aptonic.com/actions

Lots of people just use it as a quick way to open an application or folder. Not sure if you tried it before or after keyboard shortcuts were added but for example if I hit F3 then D that quickly opens my downloads folder. If I want to copy the path of the currently selected file in Finder, I use F3 then F (with the Finder path extension installed). Stuff like that. You can also select a file in Finder and then activate the Dropzone service with Ctrl+Option+Cmd+D and then have Dropzone move the file, have it uploaded or have some extension do something with it.

It can also shorten URLs. Out of the box if you select a URL and hit Ctrl+Option+Cmd+S the selected URL will be shortened using the Goo.gl URL shortening service and the new URL placed on the clipboard.

Another use, and this is an extension I've just developed recently is the ability to download videos from YouTube. With the YouTube downloader action installed, you select a YouTube video URL, hit the Ctrl+Option+Cmd+D shortcut and then press Y and the video is downloaded into the chosen folder.

Another popular Dropzone feature is the Imgur uploader integration. You just drag an image onto the Imgur action and it immediately uploads to this service and puts the URL on the clipboard. You can also use this to upload multiple images to create an album. And sure you could just do this using the Imgur website, but I would argue it's faster and more convenient to do it using Dropzone.

Also, bear in mind there are two seperate versions of Dropzone available - The Mac App Store version and the non-Mac App Store version. The Mac App Store version is a little bit crippled due to the sandboxing so it's best to use the non-Mac App Store version. It's lets you install more actions.

iCHAIT:
@aptonic Thanks for a detailed description of what all dropzone can do, I'll try it out as soon a possible. Sounds pretty useful.

aptonic:
@iCHAIT Great, thanks! 😄

herrbischoff:
Thanks for the detailed explanation. This made it a little clearer for me why exactly I did not find a use for it. Most useful functionality I already cover with other tools and as I'm a shell user most of the time, I also don't like key chords like `Ctrl+Option+Cmd+S`. I'm a die-hard Vim user, so everything more complex than `Ctrl-F` is straining my wrists... ;)

I do see the appeal for less-specialized and more GUI-focussed (aka the majority of) users though.

> You can also select a file in Finder and then activate the Dropzone service with Ctrl+Option+Cmd+D and then have Dropzone move the file, have it uploaded or have some extension do something with it.

That's what I use LaunchBar for.

> It can also shorten URLs. Out of the box if you select a URL and hit Ctrl+Option+Cmd+S the selected URL will be shortened using the Goo.gl URL shortening service and the new URL placed on the clipboard.

Again, LaunchBar or PopClip for me.

> Another use, and this is an extension I've just developed recently is the ability to download videos from YouTube. With the YouTube downloader action installed, you select a YouTube video URL, hit the Ctrl+Option+Cmd+D shortcut and then press Y and the video is downloaded into the chosen folder.

Youtube-DL.

> Another popular Dropzone feature is the Imgur uploader integration. You just drag an image onto the Imgur action and it immediately uploads to this service and puts the URL on the clipboard.

This sounds like it's interesting only if you share a lot of images publicly, which I never do except for on GitHub.
 
> Also, bear in mind there are two seperate versions of Dropzone available - The Mac App Store version and the non-Mac App Store version. The Mac App Store version is a little bit crippled due to the sandboxing so it's best to use the non-Mac App Store version. It lets you install more actions.

That's why I always get the non-MAS version if available. While I understand the need to Apple to secure its platform against technically less inclined users, I don't like to be told what an application can and cannot do. ;)

Thanks again for the examples. Since I find I cannot give a qualified recommendation one way or the other, I leave this decision in the capable hands of @iCHAIT and @dkhamsing.

iCHAIT:
> Most useful functionality I already cover with other tools.

@herrbischoff  Same here, although I think its not a bad idea to have a single tool cover all those with a set of hotkeys, though I think it is important to test the app out to see how stable it is and if it works as smoothly and efficiently as other dedicated applications do.

herrbischoff:
@iCHAIT, whatever little use I got out of it was indeed stable and worked as expected. I experienced no problems whatsoever.

raguay:
I use Dropzone all the time in my work for Envato and my personal work. I make many slide shows using a script in Dropzone that converts my markdown to a HTML slideshow. I have actions that take a zip file of ebooks, unzips them in a directory and renames each ebook according to the zip file name (they always have a numerical name only). I have another action that then repacks them. I have an action that takes a json export from a database and creates reports in pdf. There isn't a day that I do not use Dropzone, but I do have many actions I created for my own use.

PR #353, opened 2016-08-14
Add nehm

bogem:
0. [x] Project URL: https://github.com/bogem/nehm
1. [x] Why should this project be added: This app was created by myself and I think it can help for many SoundCloud users :)


iCHAIT:
Looks decent enough to me, @herrbischoff @dkhamsing thoughts?

dkhamsing:
Uh looks good, except for the downloading music part...

bogem:
@dkhamsing 

> except for the downloading music part...

What exactly?

dkhamsing:
Is this illegal?

bogem:
I think, yes

herrbischoff:
@bogem: The absurdity of the situation is quite clear. However, the fact that anyone *feels* it should be their right to do anything does not invalidate the fact that this software is in clear violation of the SoundCloud TOS (or TOU).

I propose to not include it in this list, as the only use-case is the violation of said TOS. That being said, I agree that it's decent enough software (and especially appreciate it being written in Go). People determined to find it should be able to find it without this list.

bogem:
I've understood. No problem

PR #358, opened 2016-09-01
Add yakyak

wad3g:
0. [x] Project URL: https://github.com/yakyak/yakyak
1. [x] Why should this project be added: The Google Hangouts Chrome Extension is terrible. YakYak gives you many of the same features that are offered on the native iOS/Android apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,6 @@
 
 ### Developers
 
-- [Anvil](http://anvilformac.com/) - Serve up static sites and Rack apps with simple URLs and zero configuration. ![Freeware][Freeware Icon]
 - [Base 2](http://menial.co.uk/base/) - A GUI for managing SQLite databases.
 - [Cakebrew](https://www.cakebrew.com/) - The Homebrew GUI App. [![Open-Source Software][OSS Icon]](https://github.com/brunophilipe/Cakebrew) ![Freeware][Freeware Icon]
 - [Charles](https://www.charlesproxy.com/) - Charles is a Proxy that allow you to view all of the HTTP and HTTPS traffic.
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@
 
 ### Chat Clients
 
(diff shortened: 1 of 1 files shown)


iCHAIT:
I don't think the entry is ordered alphabetically.

herrbischoff:
Another Electron wrapper application... *sigh* Therefore I repeat my stance: Electron wrappers are not awesome macOS applications.

wad3g:
@herrbischoff @iCHAIT I'm not advocating why Yak Yak should/shouldn't be added to the list, but if the reasoning for not adding it is because it's an _another electron wrapper_ why are apps like ChitChat, gInbox, and Hacker Menu on the list?

Or if it's an electron thing Atom, Slack, Nylas N1, and who knows what else.

Just curious/want clarification so I don't submit PR's for similar apps in the future.

wad3g:
K.

iCHAIT:
@wad3g You have a valid point. @herrbischoff I think we should settle this once and for all. If we do then we should include all such application or don't include any.

iCHAIT:
@dkhamsing what are your thoughts on the electron wrapped application, should they be included in the list or not?

iCHAIT:
Personally I feel they fit into the list because of the utility these apps provide, as in, a user is not expected and restricted to open the browser and use these services, many people (including me) like to use the Desktop GUI for such use cases.

dkhamsing:
Are all electron apps bad? 
I suppose it should be based on merit, so probably be on a case by case basis

If it can answer the question we have in the pull request template "Why should this project be added", that is a good start

iCHAIT:
@dkhamsing Exactly, we should treat an application on merit.

herrbischoff:
In that case, there will always be people who feel that a natively bundled web app is great and in the end most of those will be added. For me, all of the currently listed Electron apps are not really worth their salt with the one exception being Atom, as it's a hackable code editor.

PR #359, opened 2016-09-09
Remove freeware icon from Deliveries

vitorgalvao:
[Not free](https://itunes.apple.com/pt/app/deliveries-a-package-tracker/id924726344).
image

Did not fill the template since it does not apply here (edit, not an addition).


dkhamsing:
thx!

PR #366, opened 2016-09-25
Add HyperTerm to Terminals

soul-wish:
0. [x] Project URL: https://github.com/zeit/hyperterm
1. [x] Why should this project be added: 

> HyperTerm is an awesome terminal built on web technologies. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,7 @@
 
 ### Terminal
 
+- [HyperTerm](https://hyperterm.org/) - A terminal built on web technologies [![Open-Source Software][OSS Icon]](https://github.com/zeit/hyperterm)
 - [iTerm 2](https://www.iterm2.com/) - A terminal emulator. [![Open-Source Software][OSS Icon]](https://github.com/gnachman/iTerm2) ![Freeware][Freeware Icon]
 - [TotalTerminal](http://totalterminal.binaryage.com/) - A system-wide terminal available on a hot-key. ![Freeware][Freeware Icon]
 


iCHAIT:
LGTM

herrbischoff:
This is:

a) an Electron app
b) far less capable than Terminal.app itself
c) alpha quality software

I tried HyperTerm along with the other Electron terminal called [Black Screen](https://github.com/shockone/black-screen). Apart from their flashy features like completion, both terminal emulators are significantly inferior to Terminal.app. Unless at the very least `b)` and `c)` change, I vote against including it. You know my opinion of Electron apps. Plus, I don't get why you would want to implement a terminal emulator in JavaScript to run it **locally**... Darn hippies! 💇

😄

soul-wish:
@herrbischoff so Nylas N1, Atom and others in this awesome list are ok (they are also build with Electron)? 😉

dkhamsing:
Can you answer this question

Why would you use this over Terminal or iTerm

Seems like this is very slow

herrbischoff:
@soul-wish: Apart from Atom I don't feel any other app belongs on this list but they were added before I collaborated. My opinion does not reflect the teams'. Besides, please read my comment again carefully.

@dkhamsing: Good question.

soul-wish:
@dkhamsing I'm web developer, so these apps are just "closer" to me. I can create my own plugins, themes and extensions just with my knowledge in CSS and JS. I'm not sure if they are really so slow (I didn't notice it in my HyperTerm daily usage). Surely, Electron has it's own problems with perf but sometimes you can live with them (my main editor is Atom just because it is much more extendable than Sublime Text (I was a Sublime Text user for 3 years before and I still use it in case I need to work with a really big files)). 

@herrbischoff sure, everyone has their own opinion and this is the right thing. And I'm not forcing you to merge this PR (thats why it called Open Source 😉). I just want to make it clear that many really awesome products were build on the top of Electron (that was my comment about) and decided to share one more with people. Anyway, thanks for such a great list and all your work!

iCHAIT:
@herrbischoff @dkhamsing So should we close this?

herrbischoff:
I'm going to go ahead and close it. If you feel it should not be, please reopen. Perhaps when HyperTerm has matured enough to be as useful as Atom is, we can add it. However, I still feel strongly about the ill-considered notion that everything that *can* be written in JS *should* be written in JS. I'm off now learning some more ObjC and Swift to put my money where my mouth is. ;)

iCHAIT:
Hey @herrbischoff  on a different note can you take out some time and open a PR removing all electron wrapped applications from the list, in my opinion we should treat an application on its merit, having said that your argument on electron wrapped apps does not sounds entirely wrong to me. So maybe we can discuss that issue on a separate PR where we would be able to single out each application and discuss it?

If not possible for you, I can go ahead and open a PR for this myself, just let me know :)

herrbischoff:
@iCHAIT, I'm getting right on it a little later in the day.

iCHAIT:
Thanks :+1:

PR #368, opened 2016-09-27
Remove Electron wrapper apps

herrbischoff:
To discuss the merits of those applications despite the fact that they are basically web apps.

Seeing the final list of removals, I'd vote to just merge the whole PR, thereby extinguishing all traces of Electron and Chromium wrappers. Contrary to my former statement that Atom should be included, I feel that the performance gain alone of native apps like Sublime Text or even Vim justify that. In my view, none of the apps in this pull request are outstanding applications without equal or better native alternatives. Case in point: Caret vs. Typora, Atom vs. Sublime Text.

Still, I'm quite surprised to see practically all so-called "modern" editors are built using Electron. I understand the popularity and appeal. However, on a **macOS** awesome list, none of those should have a place.

Having said that, I would of course be open to also remove any cross-platform applications with no clean implementation of the Apple Human Interface Guidelines, meaning: which may be free but look like crap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,6 @@
 - [Colloquy](http://colloquy.info/) - A GUI IRC Client. [![Open-Source Software][OSS Icon]](https://github.com/colloquy/colloquy) ![Freeware][Freeware Icon]
 - [Goofy](http://www.goofyapp.com/) - Unofficial Facebook Messenger client. [![Open-Source Software][OSS Icon]](https://github.com/danielbuechele/goofy) ![Freeware][Freeware Icon]
 - [LimeChat](http://limechat.net/mac/) - Instant messaging application. [![Open-Source Software][OSS Icon]](https://github.com/psychs/limechat) ![Freeware][Freeware Icon]
-- [Messenger for Desktop](https://messengerfordesktop.com/) - An app for Facebook messenger. [![Open-Source Software][OSS Icon]](https://github.com/Aluxian/Facebook-Messenger-Desktop) ![Freeware][Freeware Icon]
 - [Profanity](http://www.profanity.im) - A terminal XMPP client. [![Open-Source Software][OSS Icon]](https://github.com/boothj5/profanity) ![Freeware][Freeware Icon]
 - [Slack](https://itunes.apple.com/in/app/slack/id803453959) - A messaging app for teams. ![Freeware][Freeware Icon]
 - [Telegram](https://itunes.apple.com/us/app/telegram/id747648890?mt=12) - A messaging app with a focus on speed and security, it’s super fast, simple and free. [![Open-Source Software][OSS Icon]](https://github.com/overtake/telegram) ![Freeware][Freeware Icon]
--- a/README.md
+++ b/README.md
@@ -132,7 +132,6 @@
 ### Editors
 
 - [Atom](https://atom.io/) - The hackable text editor. [![Open-Source Software][OSS Icon]](https://github.com/atom/atom) ![Freeware][Freeware Icon]
-- [Brackets](http://brackets.io/) - A modern text editor that understands web design. [![Open-Source Software][OSS Icon]](https://github.com/adobe/brackets) ![Freeware][Freeware Icon]
 - [Emacs](https://www.emacswiki.org/emacs/EmacsForMacOS) - Port of Emacs to work as an macOS app. [![Open-Source Software][OSS Icon]](https://emacsformacosx.com/download/emacs-sources/) ![Freeware][Freeware Icon]
 - [Light Table](http://lighttable.com/) - A customizable editor with REPL support for Clojure, JavaScript, Python and others. [![Open-Source Software][OSS Icon]](https://github.com/LightTable/LightTable) ![Freeware][Freeware Icon]
 - [MacVim](https://github.com/macvim-dev/macvim) - Vim, the text editor. [![Open-Source Software][OSS Icon]](https://github.com/macvim-dev/macvim) ![Freeware][Freeware Icon]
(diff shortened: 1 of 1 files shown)


knpwrs:
FWIW, this won't automatically apply to all Electron apps in the future. I'm currently working on an Electron app that uses native macOS APIs -- Electron is just providing the user interface.

herrbischoff:
@knpwrs: This is a good contribution to the discussion. As the base assumption here is to judge the applications based on merit (despite my own views), this is very likely to happen. I don't like those apps as I prefer to use snappy, native and consistent applications. The current fad of app-ifying everything does not sit right with me as the create-once-run-everywhere fad did years before. The best experience was, is and will always be a native application with familiar UI elements for the respective platform, be it macOS, iOS, Android, Windows, Linux (well...), whatever. There's a reason Apple dropped skeudomorphism in favour of a clean, elegant and meaningful UI. My view about UI appears to be quite at odds with yours — for me, there is no such thing as "just the user interface". As far as the user is concerned, *the interface* is *the application*.

iCHAIT:
@dkhamsing @dufferzafar Your thoughts?

knpwrs:
@herrbischoff Just curious, what would you say about applications such as Audacity or Blender (both on this list)? Audacity is WX, so it uses native widgets, but it hardly fits in with other apps (visually). Blender uses its own UI toolkit and AFAIK doesn't implement any macOS-specific functionality.

dufferzafar:
Considering this list is for macOS, I agree with @herrbischoff :+1:

akz92:
I don't think that this should be a list of the apps that have the most Apple-like UI and UX. In my opinion this should be a list of the best software available for the platform. To use text editors as an example, yes Sublime is native and performs better than Atom but Atom is still a good text editor and as such should have it's place on this list. The same applies to a bunch of 'Electron wrapper apps'. 

I believe that there's a difference between web-apps wrapped as a desktop app (Slack i.e.) and apps built on Electron (Atom i.e.).

knpwrs:
@akz92 You could argue that Slack actually does use native features not available in the web app. Two examples:

1. You can close the window and still get notifications and messages and whatnot.
2. The Electron app uses notification features that aren't available in web notifications, e.g., being able to reply directly in a notification.

Other than that, yeah, Slack is primarily a web app -- it functions almost the same in a browser whereas Atom will not function outside of Electron.

iCHAIT:
@herrbischoff Since you have got a +1 from one of the list maintainers (@dufferzafar), feel free to merge this PR, though you may want to add a line about the policy that we are going to follow for electron wrapped apps in `contributing.md` so that its a bit easier for us to reject a PR for such apps in the future and settle this argument one and for all.

vassyz:
Atom is slow, but for me it's the best editor I can get on macOS, and I've tried them all. Also Slack has got a huge thumbs up from the tech community. I agree with not having web-wrappers in this list, but banning software just because it uses a certain technology seems a bit extreme.

dkhamsing:
Of all these, I actually use Atom and Slack so I would vote to keep em 😅

Can't quite articulate why though.. just habit?

- Atom is visually very pleasing and fast enough for my use
- Slack works great and I believe the app has more native shortcuts vs the web version

lordgiotto:
I'm really sad about this PR :(
Any app should, in my opinion, be judged on it's quality and features, not on the technology behind its implementation. The technology is only the tool, but are shape and functionality that make an app awesome.
My 2 cents ;)

dkhamsing:
should we put this to a vote, whats the status

knpwrs:
_My_ opinion:

There is an [Awesome Electron](https://github.com/sindresorhus/awesome-electron) list. Things on this list would be best to show a specific strength of macOS. But that doesn't only apply to Electron Apps -- as I've mentioned earlier, Audacity and Blender on are this list but hardly show off anything about the platform or fit in with the rest of macOS. Those (and a number of Electron apps) would probably be better off on an Awesome Crossplatform list.

Maybe another section (or even another file) in this same repository would be appropriate? Awesome stuff that is available on macOS that isn't specifically awesome because of macOS?

iCHAIT:
@knpwrs Thanks for the idea mate.

@herrbischoff @dkhamsing Merging this, and for future lets encourage contributors to target their electron wrapped software suggestions to awesome electron list rather than platform specific one.

knpwrs:
@iCHAIT What would you say about the latter part of my comment? Non-Electron apps that aren't really great fits for the macOS ecosystem?

iCHAIT:
@knpwrs 

> Awesome stuff that is available on macOS that isn't specifically awesome because of macOS

I am not entirely sure to be honest, I think it is inevitable that such apps should fall in a some awesome list because they **are awesome** at the end of the day, maybe/maybe not this list.

Though, I don't think having a separate section for such apps in this list itself would harm anyone,

What say @herrbischoff @dkhamsing ?

dkhamsing:
Seems simpler for now to not allow Electron apps

If we have missed any, feel free to open a pull request

knpwrs:
I was more referring to things like Blender or Audacity. Do they belong on this list?

dkhamsing:
@herrbischoff can you take a look

iCHAIT:
@knpwrs I would suggest you to open a PR removing them so that it is more easier to scrutinize :)

herrbischoff:
@dkhamsing: I will check them out.

herrbischoff:
Please see #372.

hzb:
Forked -> Add something -> Submitted a PR -> Read the document `contributing.md` -> Cancelled submission and Deleted the forked repo.

This is my process of contributing to this repo.

PR #376, opened 2016-10-19
Add Inkdrop to Developers section

noradaiko:
0. [x] Project URL: https://www.inkdrop.info/
1. [x] Why should this project be added: A good alternative to Evernote for taking notes about programming

Thank you for making great curated list!!


iCHAIT:
Looks good

herrbischoff:
I'm voting 👎:

- it does not allow for a trial without an account
- it probably is an Electron app
- it is basically just another Markdown editor
- a coding tool that needs a continued paid subscription to be usable is bad form
- especially for something like notes
- there already is a "pay once and keep it forever" alternative on the list: Quiver

iCHAIT:
> It does not allow for a trial without an account

This is particularly odd, I thought all apps allowed a trial version without having to sign up or sign in :/

> It is basically just another Markdown editor

@herrbischoff Do you think we should have a separate sections for Markdown editors as there are a few of them in the list already?

> a coding tool that needs a continued paid subscription to be usable is bad form

This is a negative point indeed.

herrbischoff:
I believe it does not allow for a trial without account because it is a web app. It probably requires a connection to even work.

As for the Markdown editors, I feel we should rather concentrate on the best of them and leave out the rest. I propose to group all of them along with the code editors under something like "Text Editors".

PR #372, opened 2016-10-06
Non-native

herrbischoff:
This pull request removes all applications that are not truly native macOS applications.

PR #383, opened 2016-10-30
added polyphone soundfont editor

clsource:
0. [x] Project URL: http://polyphone-soundfonts.com/en/
1.  [x] Why should this project be added: Helps creating sound fonts for music creation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,9 @@
 
 - [Audio Hijack](http://www.rogueamoeba.com/audiohijack/) - Record audio from any application like iTunes, Skype or Safari, or from hardware devices like microphones and mixers.
 - [BackgroundMusic](https://github.com/kyleneideck/BackgroundMusic) - Record system audio, control audio levels for individual apps, and automatically pauses your music player when other audio starts playing and unpauses it afterwards. [![Open-Source Software][OSS Icon]](https://github.com/kyleneideck/BackgroundMusic) ![Freeware][Freeware Icon]
+- [Polyphone](http://polyphone-soundfonts.com/en/download) - A soundfont editor for quickly designing musical instruments. [![Open-Source Software][OSS Icon]](https://sourceforge.net/projects/polyphone/files/polyphone%20releases/)![Freeware][Freeware Icon]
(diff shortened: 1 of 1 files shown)


herrbischoff:
screen shot 2016-10-30 at 15 32 23

@clsource: Seriously? This is your idea of "awesome"?

clsource:
I think awesome is more than how it looks.
How helpful is a consideration too when using the awesome adjective.

This software helped me making awesome and beautiful music in many occasions.
Is a shame this awesome list is curated by such simplistic criteria.

dufferzafar:
@clsource I agree that looks aren't everything, but the thing is - I don't know much about the problem domain (music, tunes etc.) either, so being able to decide how useful the software really is tough. I'm not sure if we have a similar application on the list to just compare features.

herrbischoff:
> I think awesome is more than how it looks.

This is a given. However, if we let software like this on the list, we would also have to add a whole bunch of other, quite hideous applications. Applications which we recently summarily removed: #372 For a couple of those it could be argued to that they add really good value. Still, while they may be regarded as "awesome" in one way or the other, what they are *certainly not* is awesome macOS applications.

I don't see any reason to backtrack on that decision.

clsource:
I'm ok with your decision.
If you want more applications for music and audio making in the list.
here are some. 

http://www.reaper.fm/
http://littlegptracker.com/
http://www.deflemask.com/

cheers.

herrbischoff:
Thanks for the other links. I have checked them out and feel they are quite interesting. However, the one that stands out is DefleMask. A four-channel tracker! I haven't seen one of those since my early PC days (FastTracker) and still fondly remember NoiseTracker from my Amiga days. I cannot see a place on this list though. Maybe (probably) there's an Awesome List for audio applications out there which would no doubt benefit from your recommendations.

Reaper appears to be a cluttered mess compared to other commercial entries like Logic Audio and Cubase — both I have worked with professionally. I say that in the same way as Audacity compares to WaveLab or Mixx to Traktor. They are all without a doubt notable undertakings but simply cannot compare with the polish and feature set of their big brothers.

PR #393, opened 2017-01-06
Add Vaunt

Somnibyte:
0. [x] Project URL: http://vaunt.io/
1. [x] Why should this project be added: There's nothing like this app on your list. 😉


herrbischoff:
The reason we have nothing like this on the list is that most pro-level applications have this kind of functionality built-in. Like the myriad of available currency converters, this is clearly an application done as a learning exercise created by a novice coder. That's all fine and well (we all have to start somewhere) but it is not awesome. There's no real need for it to exist beyond personal learning. Additionally, the application itself is not especially well made.

PR #394, opened 2017-01-08
Add Lacona launcher

vinhnx:
0. [x] Project URL: https://www.lacona.io/
1. [x] Why should this project be added: in my opinion, it's better than Alfred. It's featured rich and it can process natural language. The best part, it is open source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,7 @@
 - [Instant Translate](https://insttranslate.com/mac) - Translate speech and text between 100+ languages from the menu bar.
 - [Karabiner](https://pqrs.org/osx/karabiner/) - A powerful keyboard customizer. [![Open-Source Software][OSS Icon]](https://github.com/tekezo/Karabiner) ![Freeware][Freeware Icon]
 - [Keytty](http://keytty.com) - Enables you to control your mouse with a few key strokes. Mouse Keys Alternative.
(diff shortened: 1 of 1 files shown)


herrbischoff:
Thanks for your suggestion. A couple of thoughts:

- Parts of it are open source, the full application ("Pro") is commercial. The website weirdly does not state which advantages a "Pro" purchase brings. I believe this may be due to a current beta phase.
- It appears to be written completely in JavaScript. This enables easier participation but moves it dangerously close to an Electron application.
- As far as I can see it does nothing that LaunchBar or Alfred already do. Natural language processing may be nice to have though.
- The "Pro" version is more expensive than any of the former mentioned.

I'd like to see some rather glowing support and highlighting of distinguishing features before this gets my vote for addition.

iandol:
The website says: "use of addons incur a brief delay until purchase of Pro", so all features should work for free with performance hit (unlike Alfred). But then **all** features work in Quicksilver, and that is both fully open-source and totally **free**. Perhaps when Lacona is a bit more mature, but I see no compelling reasons to use it personally so far (and I consider a good launcher possibly the most important software I use).

herrbischoff:
> use of addons incur a brief delay until purchase of Pro

I consider this unacceptable. Therefore I'm closing this for the time being. We can possibly revisit it when it is more mature and has a competitive pricing.

PR #402, opened 2017-02-06
Add Trym

LisaDziuba:
0. [x] Project URL: http://kontentapps.com/trym
1. [x] Why should this project be added: it's simple and useful app for viewing, optimizing & converting SVG icons.


herrbischoff:
This appears to be another one of those "developer is still learning" applications. The idea behind the app seems alright until you consider the target audience, which is practically zero. Graphic designers have professional tools for that task and developers are not going to pay for what boils down to a SVG-to-PNG converter. This can be done through ImageMagick with one command:

```bash
convert input.svg output.png
```

As a free tool I would consider it, not so for $4.

PR #404, opened 2017-02-15
Add Magnet

Dedados:
0. [x] Project URL: https://itunes.apple.com/us/app/magnet/id441258766?mt=12
1. [x] Why should this project be added: It is a very useful Window Management app with a great feature allowing windows to be dragged to the edge of the screen to expand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@
 - [Phoenix](https://github.com/kasper/phoenix) - A lightweight window and app manager scriptable with JavaScript. [![Open-Source Software][OSS Icon]](https://github.com/Hammerspoon/hammerspoon) ![Freeware][Freeware Icon]
 - [Spectacle](https://www.spectacleapp.com/) - Easily organize windows without using a mouse. [![Open-Source Software][OSS Icon]](https://github.com/eczarny/spectacle) ![Freeware][Freeware Icon]
 - [Stay](https://cordlessdog.com/stay/) - Resize/position windows when displays change.
+- [Magnet](https://itunes.apple.com/us/app/magnet/id441258766?mt=12) - Allows dragging a window to the desktop's edge to make it into a tile; multiple tile sizes and shortcuts.
 
 ### Others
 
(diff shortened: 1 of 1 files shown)


dkhamsing:
How much is it

Dedados:
0.99 Euro.

herrbischoff:
You can get all of this in the free and open source Spectacle. There's also BetterSnapTool, Amethyst, kwm, … What sets Magnet apart from those, not counting being popular in the Mac App Store?

herrbischoff:
BTW: "It is a great addition to this list, making it more useful!" is not a valid reason. This is highly subjective and offers no concrete example for **what it is** the app does great, especially what it does **better** than other entries in its field. Please elaborate.

Dedados:
I didn't phrase my reason well. The app is nothing special, it virtually does the same as other apps but I saw it wasn't in the list and thought why not include it, it will make the list more useful. That is why I said: 'it is a great addition to the list, making (the list) more useful'.

dkhamsing:
Closing
Thanks for the suggestion

Dedados:
Why not include it?

herrbischoff:
Because this is an **Awesome List**. We strive to include only the best applications of any kind. In your own words:

> The app is nothing special, it virtually does the same as other apps but I saw it wasn't in the list and thought why not include it

This is decidedly not the definition of being "awesome".

Dedados:
Ok, thanks for clarifying your reasoning:)

herrbischoff:
Thanks for being understanding. :)

lordgiotto:
My 2 cents: I've used both Spectacle, Better Snap (and Touch) Tools and Magnet.

With its small .99 cents cost, Magnet is really worth it! 

I found it to be less buggy than better snap tools (which costs 3 times more); compared to Spectacle it gives the same powerful keyboard shortcuts and the drag and drop edge resize.

So, in my opinion, it's more *awesome* than the mentioned alternatives 😉

Dedados:
Since I have never used others, I cannot comment on them but I think magnet is worth it as it works flawlessly. It is also the only tool I know of (other than moon) which gives you the ability to resize by dragging to the edge of the screen.

Dedados:
In fact, I just checked and it is also 10x cheaper than moom despite doing the same job. Moom has a few more tools but that makes it unnecessarily complicated for the things I and expect others would do.

dkhamsing:
Thx for the endorsement
Let's have @iCHAIT weigh in

iCHAIT:
I use spectacle and it works pretty good for me so haven't felt the need to look for its alternatives, haven't used Magnet so can't comment much on that, however @lordgiotto endorsement of the app seems pretty legit to me. Maybe we can reopen this and place a **Need Endorsement** label on it and if we get another +1 we add it to the list, does this sound good?

Dedados:
Good idea.

dkhamsing:
let's close this, lack of endorsements 😅

Dedados:
Seems like there wasn't 'attraction'.

PR #410, opened 2017-03-13
Add Boom 2

lordgiotto:
0. [ ] Project URL: http://www.globaldelight.com/boom/
1. [ ] Why should this project be added: As developer, immersive music is an essential work tool, like the text editor. Since I started using Boom2, I can't live without it anymore. Audio enhancement is really awesome and the difference with default equalization is huge. Futhermore MacOS doesn't have a built-in audio equalizer.
2. [ ] End all descriptions with a full stop/period
3. [ ] Entry is ordered alphabetically
4. [ ] Appropriate icon(s) added if applicable (OSS, freeware)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@
 
 - [Audio Hijack](http://www.rogueamoeba.com/audiohijack/) - Record audio from any application like iTunes, Skype or Safari, or from hardware devices like microphones and mixers.
 - [BackgroundMusic](https://github.com/kyleneideck/BackgroundMusic) - Record system audio, control audio levels for individual apps, and automatically pauses your music player when other audio starts playing and unpauses it afterwards. [![Open-Source Software][OSS Icon]](https://github.com/kyleneideck/BackgroundMusic) ![Freeware][Freeware Icon]
+- [Boom 2](http://www.globaldelight.com/boom/): System wide equalizer and audio enhancer, with volume boost, auto equalization, presets and variuos audio effects.
 - [VOX Player](https://coppertino.com/vox/mac) - Play numerous lossy and lossless audio formats.
 - [XLD](http://tmkk.undo.jp/xld/index_e.html) - Rip/Decode/convert/play various "lossless" audio files. [![Open-Source Software][OSS Icon]](https://code.google.com/archive/p/xld/source) ![Freeware][Freeware Icon]
 


dkhamsing:
Starts at $15

herrbischoff:
I tried it some time ago and deleted it after a week of sporadic usage. It promises more than it delivers and is mostly focused on improving the sound of the built-in speakers. I found Audio Hijack in live mode to be a far more flexible approach to improve any kind of sound if needed. To me, this is almost crapware. It does one thing and not very good. So, thumbs down from me.

lordgiotto:
@herrbischoff They has different user targets: Audio Hijack costs 49$ and seems to be a very advanced and feature rich tool. Boom2 is more a consumer and it-just-works solution, imho.

However the sound quality improvement is really noticeable to me (it's one of the few apps I decided to buy for MacOS after the trial period), but maybe I'm not enough audiophile to tell the difference with a professional tool :)

herrbischoff:
As the largest difference of Boom is experienced through the built-in speakers, you are probably listening to music through them while at the same time stating "immersive music" as being indispensable to you. Huh?

If you listen to any kind of local music and need an equalizer, just about every single music player ever built has one: iTunes, Vox, you name it. Free.

If you listen to streaming services like Spotify, expecting quality sound, you cannot be helped. Seriously. The compression distorts the sound in a way that post-processing is putting make-up on a pig. Streaming services are in the content-delivery business where the size of the data delivered translates directly into cost.

Bonus point, as you are concerned about immersion: as lossy audio compression works by eliminating data and masking it, your brain has to reconstruct the missing parts. There's no compression without decompression. The difference is, that in this scenario, your brain does the decompression. While it is quite skilled at that, it also needs to work more than it would normally have to. Listening to compressed music formats taxes your brain more, eventually affecting your concentration.

Also, expectation and perception are big contributors to how sound is experienced. Slap a little more bass and mid-range on and most people will say it sounds significantly better than before.

I would not consider a tool to change how my internal speakers sound to be worth $15 unless it yielded some rather spectacular results, which is physically impossible to do with small speakers.

lordgiotto:
@herrbischoff I understand your point, and thanks for the "lossy audio thing", I didn't know it :)

At the end of the day, the key difference between us is that I'm not an audiophile: I use Apple Music/Spotify and I listen music with in-ear urBeats. 
I don't care about audio fidelity (probably I couldn't tell the difference), but I care about immersiveness, I like high basses and mid-ranges and i would't be able to equalize manually my music.

To me (and to people like me, I'm sure I'm not the only "ignorant" music end-user) Boom is awesome, thanks to its out-of-the-box equalization, its straightforward and simple interface and its "oh, it sounds really good now" feeling when you activate it :) 
And I paid it 13$ for 2 user licenses.

However I really get your arguments, I would react the same way if someone proposed me some paid git tool over the powerful bash, or some trivial photo editor over versatile tools like Affinity Photo. But, probably, they are different "targets", like we are in this case ;)

dkhamsing:
Closing for lack of endorsement

PR #415, opened 2017-03-16
Add Touch Bar Simulator

LisaDziuba:
0. [x] Project URL: https://github.com/sindresorhus/touch-bar-simulator
1. [x] Why should this project be added: it makes easier to code and design for the Touch Bar (if person doesn't want or can't launch the simulator from inside Xcode)


LisaDziuba:
I found this tool on Product Hunt today. I haven't work with apps that use Touch Bar yet, but the idea of having simulator for this is very logical. Touch Bar Simulator requires macOS 10.12.2 or later, that's should be noted also.

I see it useful for QAs folks and UX & UI designers, who might use this simulator to test Touch Bar support (no need to install Xcode). 

Check it :)

dkhamsing:
@herrbischoff any thoughts

PR #417, opened 2017-03-22
Add FinderGo

onmyway133:
0. [x] Project URL: https://github.com/onmyway133/FinderGo
1. [x] Why should this project be added: It helps open terminal from Finder, which we  need 100 times a day

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@
 - [Quicklook-Plugins](https://github.com/sindresorhus/quick-look-plugins) - List of extra quicklook plugins to enable previewing more filetypes in the Finder.
 - [TotalFinder](http://totalfinder.binaryage.com/) - A powerful alternative to Finder.
 - [XtraFinder](https://www.trankynam.com/xtrafinder/) - Adds useful features to Finder. ![Freeware][Freeware Icon]
-
+- [FinderGo](https://github.com/onmyway133/FinderGo) - Open terminal quickly from Finder [![Open-Source Software][OSS Icon]](https://github.com/onmyway133/FinderGo) ![Freeware][Freeware Icon]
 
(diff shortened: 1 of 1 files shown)


herrbischoff:
This application is a pointless and convoluted way to do simple one line commands inside Terminal to achieve the same. To do the same from terminal is even easier: `open .`

onmyway133:
@herrbischoff Hi, you may read it wrong. `open .` from Terminal will open Finder. While this allows us to open Terminal from Finder

herrbischoff:
Nope. [This](https://github.com/herrbischoff/awesome-osx-command-line/blob/master/README.md#change-working-directory-to-finder-path) is all you application does.

onmyway133:
@herrbischoff I think you have your point. Thanks for your time

herrbischoff:
Sure. It's a nice learning project to get acquainted with Swift but unfortunately does not qualify for this list.

iCHAIT:
@herrbischoff I get where you are coming from. However, the fact that this is fairly popular within the community, makes me think twice.

Also since this is an awesome list, I think we should add software that is indeed awesome and mac oriented (which this one is). I understand that what this app does can be done via a shell command, but I have seen people using such apps. I think it would definitely be useful for some of the people out there.

@dkhamsing what do you think?

herrbischoff:
Comment on wrong PR? ;)

Yeah, that's exactly why I'm now standing neutral on this. I can see people using something like this even if I wouldn't.

iCHAIT:
> Comment on wrong PR?

Haha, sorry my bad.

Yeah, so I'll reopen this one as well? And put an endorsement tag on both (this one and #521)

PR #429, opened 2017-05-13
Add SpaceLauncher

oolugh:
0. [x] Project URL: https://spacelauncherapp.com/
1. [x] Why should this project be added:

Before macOS Sierra, I used to launch or switch to an app with [Brett](http://brettterpstra.com/contact/)'s ["hyper key" configurations](http://brettterpstra.com/2012/12/08/a-useful-caps-lock-key/). It remaps Caps Lock key (called "hyper key") to Command+Shift+Option+Control via [Karabiner](https://pqrs.org/osx/karabiner/), and use "hyper key" + another key to launch or switch to an app. It saved me a lot of time.

However, Karabiner is not compatible with macOS Sierra. After the upgrade, I really missed the configuration. Without "hyper key", I had to use Dock, Command+Tab, Launchpad or Mission Control to switch apps. I started to notice I was distracted with locating the target app I want to switch to.

SpaceLauncher lets users launch or switch to an app by holding down the spacebar while pressing a key. Spacebar is so easy to be pressed and combined with another key to switch apps. This makes my workflow much smooth and less distraction. I hope others can also benefit from SpaceLauncher.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,7 @@
 - [Quicksilver](https://qsapp.com/) - Control your Mac quickly and elegantly. [![Open-Source Software][OSS Icon]](https://github.com/quicksilver/Quicksilver) ![Freeware][Freeware Icon]
 - [SelfControl](https://selfcontrolapp.com/) - Block access to distracting websites. [![Open-Source Software][OSS Icon]](https://github.com/SelfControlApp/selfcontrol/) ![Freeware][Freeware Icon]
 - [Simplenote](https://simplenote.com/) - Simple cross-platform note taking app with cloud-based syncing. ![Freeware][Freeware Icon]
+- [SpaceLauncher](https://spacelauncherapp.com/) - Hold down spacebar while pressing a key to launch or switch to an app.
 - [TaskPaper](https://www.taskpaper.com/) - Plain text to-do lists.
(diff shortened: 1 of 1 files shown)


herrbischoff:
@oolugh: The project URL is missing in the description.

Apart from that: $14.99 is just unacceptable for this kind of application. If it were free or maybe $1-2, I'd endorse it. Personally, the space key setting would interfere with my Vim setup using space as the leader key. This is an edge case however. Still, this is in "useful helper" territory, doing one thing only (which is good). I would never pay for it, let alone $14.99.

oolugh:
@herrbischoff Sorry, I can't believe I gave detailed reason but forgot the URL. I updated.

oolugh:
@herrbischoff About the price, SpaceLauncher doesn't stop any functions after trial expiration. It seems only prompt an alert after hundreds of uses. So I guess the price is just a kind of donation.

gviridis:
I love it! And the current price is $4.99!

herrbischoff:
> I love it! And the current price is $4.99!

It wasn't when I checked the page. The current pricing is much closer to the mark in my opinion. Let's see what the others say.

In any way, good to see some endorsement. And thanks to @oolugh for writing up the reasoning behind his suggestion. I suppose that's because he's the author of the app. ;)

herrbischoff:
Interesting: just noticed that the application is now completely free. That was fast. As soon as @dkhamsing or @iCHAIT review it and give the green light, we can probably merge.

PR #431, opened 2017-05-22
Add Qbserve

ivmirx:
- [x] Project URL: https://qotoqot.com/qbserve/
- [x] Why should this project be added: a time tracker designed for Mac, it stores all your data locally and doesn't require subscription (I'm the developer of the app)


vassyz:
I've been using this app since day one; I think it was a HackerNews post. I can confirm it deserves to be on this list. One big downside to it would be the price, but quality-wise, I have no complaints.

herrbischoff:
<https://timingapp.com> comes to mind as an alternative, which is what I use. It's also $10 less.

dkhamsing:
Not too familiar with this space, the app looks good

dkhamsing:
thx

PR #432, opened 2017-06-01
Add Rocket to Productvity

vikrantnegi:
0. [x] Project URL: http://matthewpalmer.net/rocket/
1. [x] Why should this project be added: Makes adding emojis faster while typing even where they are no tool to add emojis. You can blacklist app where you don't want emoji suggestions like editors.


iCHAIT:
I really like this. I tried it out as well and it works really well.

The only concern I could see is that it does not work on all applications, and I believe for some applications it does not work on purpose (for instance on code editors like Sublime Text), and other applications may not be supported. But it worked on most, and I like it.

👍  from me.

@herrbischoff @dkhamsing what do you think?

vikrantnegi:
@iCHAIT Yeah, I really like this and yes you can filter out apps that you don't want emoji's suggestions. I got to know about this from [Wes Bos](http://wesbos.com/uses) personal site. He uses it on his videos tutorials.

dkhamsing:
Looks good! Merging

herrbischoff:
OMG, emojis! Now we can all text like teenage girls, all the time! ❤️ 😍 😻 ❤️ 😍 😻 ❤️ 😍 😻 ❤️

Sorry, but I don't get this emoji thing. :P

PR #444, opened 2017-09-24
Add Developers tool

LisaDziuba:
1. [x] Project URL: https://flawlessapp.io/
Flawless App helps developers to easily spot visual differences between design and implementation right inside iOS simulator. With Flawless mobile teams can make polished and nice looking apps very fast. The tool has overlay\split comparison modes and support of long\landscape design screens comparison. 


LisaDziuba:
We made Flawless App together with Ahmed & our Ukrainian team. We spent 2 years to make, test and polish it. Now it’s easy-to-use well-tested tool with full support. 

We want to make Flawless App accessible for developers with a different financial situation, that's why we put a fair price for it ($15 for life-long use). Beta-version was free and it will be free for early users forever. I know, that community prefer OS or free products, as well as we do. But, like many tools, we have to go to commercial product for continue improving it without investments :)

herrbischoff:
@LisaDziuba: I'm unsure about this and would have to test it to give an opinion, which is probably taking some time.

stackptr:
I tried it this morning. I don't do iOS development but this app looks very slick as a native app with menubar component, first-time user tutorial, and seems to provide a useful development and design niche. I think it should be added unless there is some missing context an iOS developer could add.

herrbischoff:
As we now have two endorsements, I'm going to merge this.

LisaDziuba:
thank you 🎉 🎉 🎉

PR #446, opened 2017-10-03
Add IconJar to Graphics

alexchantastic:
1. [x] Project URL: https://geticonjar.com/
2. [x] Why should this project be added: Icon management is a pain point for designers and this app solves it quite well.


herrbischoff:
This looks promising to me, although I would have to question the utility of what basically boils down to an icon catalog with export functionality at a price of $30. You have to ask yourself how often you truly need access to an overview like this as opposed to a clearly named folder of SVGs and a QuickLook plugin. The drag-and-drop function is neat though but can be easily replicated at no cost in Finder, the exception being icon fonts. Given that every icon font I know features overviews and you'd stick to one collection for consistency anyway, combined with the fact that IconJar stores its data in an iTunes-like database (lock-in), I don't see real value here.

herrbischoff:
Since no one else is objecting, I'm going to close this.

PR #456, opened 2017-10-15
Add gitX dev

LewisVo:
1. [x] Project URL: https://github.com/rowanj/gitx
2. [x] Why should this project be added: Defacto standard GUI for Git on Mac. Even recommended on Git home page https://git-scm.com/download/gui/mac


herrbischoff:
Do you truly think this a best-of-breed application, considering free alternatives like GitUp or even SourceTree? To me, GitX has always been an outdated, cluttered application that never fully recovered. What are the strengths that eclipse its alternatives?

LewisVo:
It is one of the best, in my humble opinion. Try clone a Github repo with Sourcetree and it fails instantly.  GitUp is definitely cleaner in terms of user interface, but GitX-dev shows everything at opening stage. You dont have to tap space, click on a tree's node to view what the commit message it, like GitUp. It's all right there, in front of your eyes with the some effort of scrolling. The blame, the history.... everything is just a tap. And all that happens with a decent amount of ram and cpu usage (if you are gonna compare with fancy things like Github Desktop, GitKraken, etc. Electron apps). It only falls short to GitUp in the code review section because GitUp has a split panel while GitX doesn't.
To sum up, I think it deserves a place here. It is also free, even open sourced.

herrbischoff:
As an alternative to the other applications I believe this may be to the taste of others, even if it's not to mine.

PR #471, opened 2017-12-19
Add TablePlus

dck:
1. [x] Project URL: https://tableplus.io
2. [x] Why should this project be added: This is a native application supporting many databases. It has a very good and lightweight UI, syntax highlighting, autosuggestion (SQL syntax, tables, columns), he history and partial code execution (what you've selected). 

However, I'm not quite sure about "Free" icon. There is paid plan but it's totally ok to use a free option for personal purposes. Paid plan suggests themes, more than 4 connection, etc.


herrbischoff:
Looks interesting. However, the asked price is steep, especially considering that you would enter into a subscription agreement to pay $50 every year. There is no guarantee that you will receieve $50 worth of updates because you'd be paying upfront. I generally don't like subscription models for software.

stackptr:
How do you feel the free version compares to Postico's free version (which is already added)? Postico limits to 5 connections, single window per connection, and no table filters. TablePlus limits to 4 connections, 2 tabs, 1 window, and 3 favorites. It seems that both free versions are comparable in their limitations unless I'm missing something (I use Postico's free version and have no tried TablePlus)

PR #472, opened 2017-12-27
Add SimpleumSafe

yjogol:
1. [x] Project URL: [https://simpleum.com/](https://simpleum.com/)
2. [x] Why should this project be added: New level of protection. Encrypted while viewing, editing and syncing. 


herrbischoff:
I'm not too impressed by encryption products that lock you into their platform and are commercial in nature. This basically means that security becomes attainable only for those who can shell out the money for it. As with all full-price commercial software, you are entirely at the mercy of the developer to support new OS versions, which will lead to shelling out ever more money for new versions.

I'm open to other viewpoints, this is just my opinion on the matter.

iCHAIT:
@dkhamsing thoughts?

simpleum:
Hi, we are the developers of SimpleumSafe.

I do not what to start a discussion about commercial and free apps.
I personally have used over 20 years free encryption apps, but I had to have often to change the app, because a project was discontinued.
Than we have decided to build an own encryption app with passion and added all the features we have missed in other apps.
It was a big effort to build it, so we decided to sell it, because we spend thousands of hours creating and maintaining the app.
So passion becomes a hobby and the becomes a business, so we work on what we love.

We wanted to build an easy to use encryption app which you can use without technical knowledge. 
Now we have thousands of users and they are very happy to have an easy to use encryption app.
We realized many Mac users like to have productive apps, which are easy to use without technical knowledge and give them confidence to do such important tasks like file encryption with guidance to do it in the right way.

Is this list only for free apps? or do you think this app doesn't add value for every day tasks?

Please add it to the list.

Do you have any question for the decision?

dkhamsing:
I am neutral on this. The list should only be for good apps, free is not a requirement.

iCHAIT:
I think it fits the list. Merging 👍

PR #474, opened 2018-01-08
Add link to fman (dual-pane file manager)

mherrmann:
1. [x] Project URL: https://fman.io
fman is the most modern dual-pane file manager and the first to integrate innovative features from Sublime Text. When it launched in March 2017, fman made the top 10 on Product Hunt and stayed on the front page of Hacker News for more than 13 hours.

The default spelling for fman is with a lower-case 'f'. I left it like that for that reason even though `contributing.md` says to capitalize titles. However, I would also be happy to capitalize it if you like. Just let me know please and I'll update this PR.

fman is [not based on Electron for performance reasons](https://fman.io/blog/picking-technologies-for-a-desktop-app-in-2016/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@
 
 ### Finder
 
+- [fman](https://fman.io) - Modern dual-pane file manager integrating features from Sublime Text.
 - [ForkLift](https://itunes.apple.com/us/app/forklift-file-manager-ftp/id412448059) - File Manager and FTP/SFTP/WebDAV/Amazon S3 client.
 - [Path Finder](http://www.cocoatech.com/pathfinder/) - A powerful dual-pane browser alternative to Finder.
 - [Quicklook-Plugins](https://github.com/sindresorhus/quick-look-plugins) - List of extra quicklook plugins to enable previewing more filetypes in the Finder.


herrbischoff:
This being a PyQT project that is completely sandboxed in itself my question is: what exactly is the functional difference compared to a plain shell file manager like Midnight Commander, vifm oder Ranger despite being slower and featuring some icons?

mherrmann:
One of the things that for instance sets it apart from shell file managers is a feature called _GoTo_ for quickly jumping to directories. It auto-completes paths similar to Spotlight. Being graphical rather than shell-based it fits in the same niche as Forklift, which you have on the list. fman's difference is that it is much more open through a [Python-based plugin API](https://fman.io/docs/plugins-introduction).

One of the nice things that is macOS specific is that, through the plugin API, it easily integrates with Alfred. For instance, one user has an Alfred workflow that opens the selected file in fman and opens it in Alfred Browser.

herrbischoff:
Thanks for elaborating. With regard to ForkLift, it's actually a native macOS application that integrates with the Services menu, leverages AppleScript and has the ability to define custom scripting without the need to write plugins, among many other things. The auto-completion is alright but every modern terminal features that as well. The plugin API is not macOS-specific but rather general. The Alfred example probably works through application-specific URLs, which could be replicated on any OS with any applications that support it.

fman is not my cup of tea and I don't think it's well placed here, being focused on being a cross-platform application. I'll let the other contributors decide whether to include it or not. For my taste, it does not offer a compelling enough difference to the established solutions.

mherrmann:
Okay, thank you. I look forward to hearing what the other contributors think.

iCHAIT:
👎

PR #488, opened 2018-03-20
Add Trailer to Developers section

stackptr:
1. [x] Project URL: https://github.com/ptsochantaris/trailer
2. [x] Why should this project be added: Seems to be very flexible/configurable, written in Swift, and has a feature set that is not provided by other apps in this section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@
 - [Swiftify](https://objectivec2swift.com/#/xcode-extension/) - Objective-C to Swift code converter and Xcode & Finder extensions.
 - [Touch Bar Simulator](https://github.com/sindresorhus/touch-bar-simulator) - The macOS Touch Bar Simulator as a standalone app. [![Open-Source Software][OSS Icon]](https://github.com/sindresorhus/touch-bar-simulator) ![Freeware][Freeware Icon]
 - [Tower](https://www.git-tower.com/) - The most powerful Git client.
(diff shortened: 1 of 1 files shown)


herrbischoff:
Why would you want to use a separate application just for managing issues? It appears to be rather unproductive since it's the only thing you can do with it. Additionally, I believe we're terrorized enough by notifications as is, so I really cannot see the point in getting GitHub notifications not only on my laptop but also on my phone and my watch (that I don't have). So, thumbs down from me because the whole premise of this app is flawed. Just because you **can** do something doesn't necessarily mean that you **should**.

stackptr:
I needed something that could act as a persistent reminder for PRs on work projects that I sometimes defer and then forget about. For my purposes, this is actually the opposite of workflow disruption, as it (at least in theory, I've only used it for a few days) allows me to get to a satisfying stopping point before context switching to review mode. For what it's worth, I have most notifications turned off. I can't argue against your personal preference around using this app, but I do think that it fits the criteria of being on this list and that we are not required to personally use or endorse every app on the list.

herrbischoff:
You’re entitled to your opinion and your recommendation is welcomed and will be considered. However, you should refrain from trying to explain to the actual maintainers what this list is for. To put it bluntly: you don’t get a say. If you disagree, happily start your own one.

iCHAIT:
I think this is useful 👍

PR #492, opened 2018-04-25
Fixed MPU Dead link

bxio:
The Mac Power Users link was broken, now fixed.


dkhamsing:
Thank you 

PR #494, opened 2018-04-30
GitFTP-Deploy

jannejava:
1. [x] Project URL:https://gitftp-deploy.com
2. [x] Why should this project be added: Time saver for your smaller web projects where full-on deployments are not necessary, but you still want to use git.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@
 - [Flawless App](https://flawlessapp.io/) - macOS tool for visual quality check of mobile app right inside iOS simulator.
 - [Gas Mask](https://github.com/2ndalpha/gasmask) - A simple hosts file manager which allows editing of host files and switching between them. [![Open-Source Software][OSS Icon]](https://github.com/2ndalpha/gasmask) ![Freeware][Freeware Icon]
 - [gitbar](https://github.com/Shikkic/gitbar) - Open source github contribution stats on your Menu Bar. [![Open-Source Software][OSS Icon]](https://github.com/Shikkic/gitbar) ![Freeware][Freeware Icon]
-- [GitFTP-Deploy](https://gitftp-deploy.com) - Use Git to track changes and upload only the changed files with FTP
+- [GitFTP-Deploy](https://gitftp-deploy.com) - Use Git to track changes and upload only the changed files with FTP.
 - [GitHub Desktop](https://desktop.github.com/) - A GUI for using GitHub. ![Freeware][Freeware Icon]
 - [GitUp](http://gitup.co/) - A simple but powerful Git macOS app. [![Open-Source Software][OSS Icon]](https://github.com/git-up/GitUp) ![Freeware][Freeware Icon]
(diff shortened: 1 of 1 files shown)


herrbischoff:
What's your use-case and audience for this? You can have an arguably more integrated (you do show Git committing via the shell on the website) solution like https://git-ftp.github.io — which is free and open source software, as easily installed as `brew install git-ftp`. You'd also avoid paying a rather steep €24.95 this way and don't risk the software being abandoned and incompatible with future macOS releases if unprofitable or uninteresting to the author at some time. Sadly, that's a common occurrence. I'm sure there are plenty of people valuing a GUI but in this case I'm afraid I cannot see the value in it. What exactly does it improve upon?

jannejava:
Hi Marcel!

Seems like you is not the target audience for this app. Have you experience of git-ftp? Never spent time dealing with broken homebrew, gem or pip installations at every macOS update? So common developers don't value their own time when they can save 24.95 EUR. Here everything comes in one package with support. Drag and drop and it’s installed.

I think the target audience is not the most tech savvy web developers, which are just starting out and justing git. Most of them are using something like DeployBot, DeployHQ €8.5/month for 10 repos. If you stop paying you will have nothing.

/ Jan

herrbischoff:
Nope, Homebrew never broke for me, neither did Ruby or Python for that matter. When you're doing it properly with `rvm` and `pyenv` and don't update to a new macOS version the day it comes out, your dev environment is very unlikely to break, like, ever. My point is not about saving 25 bucks but having to pay 25 bucks for a product that effectively doesn't do anything but provide a GUI. But that's just me.

Am I correct in assuming that your target audience appears to be people just starting out with web development and Git, kind of like [CodeKit](https://codekitapp.com/) is for people not comfortable with tools like [Brunch](http://brunch.io/)?

iCHAIT:
I don't see much use of the application for myself either. Homebrew, ruby and python never broke for me, it's just about using appropriate version managers properly as @herrbischoff rightly mentioned. Again that is me, other's may have different viewpoints. As @herrbischoff mentioned newbie developers may find value in this.

As far adding this to the list is concerned I am not sure since use case is pretty subjective.

herrbischoff:
A little background to my opinion: back when I started with web development, simple tools like Brunch didn't exist, let alone stuff like Webpack. Looking back, I regret that I once used isolated helper apps like CodeKit at all. They don't scale.

An extremely unpleasant experience with the testosterone-fuled ego of the CodeKit developer (including but by no means limited to using "bro-speak" right out of the gate) prompted me to take a deep-dive into Gulp (Grunt was and is more trouble than it's worth). So, in the end I was almost grateful for his rather abrasive personality. Sure, it was painful to learn at first but this eventually opened up into future venues like Webpack (which continues to be a pain by its own). We now have a wealth of task runners, bundlers and build tools available that are quite agreeable to learn and have solid starting templates. There are tutorials all around to get you up and running. I really cannot stretch enough how much easier Brunch makes building simple, low-complexity projects.

Plus, nowadays, the only reason I can think of to deploy sites via FTP is to cheap shared hosting, which will most likely be WordPress sites and most definitely be written in PHP. For such occasions there is a neat script doing all this for you ([PHPloy](https://github.com/banago/PHPloy)), which also offers rollback functionality. For everything else there is [Deployer](https://deployer.org/). FTP by itself is insecure technology (plaintext passwords) and long deprecated in most of the industry except for public software update servers and penny-hosting.

PR #495, opened 2018-05-01
Add Mailplane

FoxxMD:
- [x] I have read the [Contribution Guidelines](https://github.com/iCHAIT/awesome-macOS/blob/master/.github/contributing.md)
- [x] Project URL: https://mailplaneapp.com/
- [x] Why should this project be added: Unlike other mail clients it leaves the GApps interface intact so the user can use it as if it's in-browser but provides all the bells and whistles you'd want like notifications, sharing to other apps, etc. so it's not Just Another Mail Client ™

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@
 
 - [Airmail](http://airmailapp.com/) - Lightning fast email client designed for El Capitan.
 - [MailMate](https://freron.com/) - Advanced IMAP email client, featuring extensive keyboard control and Markdown support.
+- [Mailplane](https://mailplaneapp.com/) - A tightly integreted client for Google Mail, Inbox, Calender, and Contacts.
 
 
 ### Finder


stackptr:
FYI, this does _not_ appear to be built with Electron. No `app.ajar` [1] and the main tab content appears to a modern WebView implementation (inspected it shows Dev Tools identical to what Safari has).

1: https://stackoverflow.com/questions/42342048/how-do-i-determine-if-an-application-is-built-by-electron-or-not
2: https://developer.apple.com/documentation/webkit/webview

iCHAIT:
@herrbischoff I don't think you are, However I really don't understand what is particularly awesome about this application or how is it different from Gmail or Mac's Mail.app.

iCHAIT:
Thanks for the detailed elaboration @FoxxMD 

Looks good to me :+1:

PR #505, opened 2018-10-01
Added Commander One

jakejdavis:
0. [X] I have read the [Contribution Guidelines](https://github.com/iCHAIT/awesome-macOS/blob/master/.github/contributing.md)
1. [X] Project URL: https://mac.eltima.com/file-manager.html
2. [X] Why should this project be added: The base version offers a free dual-pane Finder alternative with the paid version offering FTP, Google Drive and Dropbox etc. integration
3. [X] End all descriptions with a full stop/period
4. [X] Entry is ordered alphabetically
5. [X] Appropriate icon(s) added if applicable (OSS, freeware)

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@
 
 ### Finder
 
+- [Commander One](https://mac.eltima.com/file-manager.html) - A free dual panel file manager that helps you manage your files in the best possible way.
 - [ForkLift](https://itunes.apple.com/us/app/forklift-file-manager-ftp/id412448059) - File Manager and FTP/SFTP/WebDAV/Amazon S3 client.
 - [Path Finder](http://www.cocoatech.com/pathfinder/) - A powerful dual-pane browser alternative to Finder.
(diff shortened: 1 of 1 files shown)


iCHAIT:
👍 from me

@herrbischoff @dkhamsing Thoughts?

herrbischoff:
We already have the excellent [ForkLift](https://binarynights.com/) on the list. From what I can tell, Commander One does nothing useful ForkLift doesn't do already and ForkLift is way more Mac-like. No need for an all-in-one solution that does many things but nothing quite right. The main difference appears to be its business model (free + pro) and missing support for things like Mojave's dark mode. Also, at comparable functionality, ForkLift is 4 cents cheaper... ;)

Seriously though, I wouldn't recommend it. So, 👎 from me.

iCHAIT:
Thank you for the detailed review @herrbischoff. I missed out the details.
Closing

PR #514, opened 2019-01-09
Added XiEditor

liamrosenfeld:
- [x] I have read the [Contribution Guidelines](https://github.com/iCHAIT/awesome-macOS/blob/master/.github/contributing.md)
- [x] Project URL: https://xi-editor.io
- [x] Why should this project be added: It is an extremely smooth text editor that can handle extremely large files with ease


herrbischoff:
> Note: This project is still in an early state. Prebuilt binaries will be made available once we start creating versioned releases.

Wait until there is a release and then we can revisit it.

PR #480, opened 2018-01-28
Replace Bookmark Utilities with News Readers category

stackptr:
0. [x] I have read the [Contribution Guidelines](https://github.com/iCHAIT/awesome-macOS/blob/master/.github/contributing.md)
1. [x] Project URL: N/A
2. [x] Why should this project be added: I wrote out my rationale in #477 
3. [x] End all descriptions with a full stop/period
4. [x] Entry is ordered alphabetically
5. [x] Appropriate icon(s) added if applicable (OSS, freeware)

PR #525, opened 2019-07-01
Update table of contents with News Readers section

stackptr:
#480 removed the "Bookmark Utilities" section and added the "News Readers" section, but the Table of Contents was not updated to reflect this change.


herrbischoff:
@iCHAIT: Also, an obvious improvement.

@stackptr: Thanks again!

PR #528, opened 2019-08-08
Add Bidbar

alankbi:
1. [x] Project URL: https://www.getbidbar.com
2. [x] Why should this project be added: It allows you to manage bash commands from the menu bar and run or copy them using auto-generated keyboard shortcuts. It acts as a faster and easier replacement/alternative to bash aliases and Automator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@
 
 - [Anvil](http://anvilformac.com/) - Serve up static sites and Rack apps with simple URLs and zero configuration. ![Freeware][Freeware Icon]
 - [Base 2](http://menial.co.uk/base/) - A GUI for managing SQLite databases.
+- [Bidbar](https://www.getbidbar.com) - Manage bash commands from the menu bar and run them with keyboard shortcuts.
 - [Cakebrew](https://www.cakebrew.com/) - The Homebrew GUI App. [![Open-Source Software][OSS Icon]](https://github.com/brunophilipe/Cakebrew) ![Freeware][Freeware Icon]
 - [CocoaRestClient](https://mmattozzi.github.io/cocoa-rest-client) - An app for testing REST endpoints. [![Open-Source Software][OSS Icon]](https://github.com/mmattozzi/cocoa-rest-client) ![Freeware][Freeware Icon]
 - [Dash](https://kapeli.com/dash) - An API Documentation Browser and Code Snippet Manager.


herrbischoff:
I don't like most about this. It looks like a junior programmers' first more polished application that "solves" a need most do not have. If you're collecting snippets in a fancy manner, you're better off with dedicated applications like [Quiver](http://happenapps.com/#quiver). You cannot run those snippets directly but I'd say there's really no need to in the first place. On top of this, it's commercial, asking $10 for some rather trivial functionality, the same as fully-features snippet managers. At least it's written in Swift.

alankbi:
Thanks for the feedback. If I may, I'll try to address some of your concerns:

**Solves a nonexistent need:**
The goal of Bidbar is to solve two specific needs of programmers:
1. The first is keeping track of all the various bash commands. For newer programmers less experienced with the Terminal (and even those more experienced who can't be bothered setting up aliases), it's easy to forget commands like resetting the latest commit, using grep to search for keywords, etc. Instead of having to Google these things over and over, Bidbar allows you to save them in the menu bar for easy access. When you need them, you can copy them with two clicks or simply use the keyboard shortcut, allowing you to paste it into your Terminal. Quiver looks great, but it seems a bit overkill and possibly not as well-suited to be used in a situation like this. 
2. The second is the running of commands. You mentioned that there isn't much need for this, but I think that it can be useful on many occasions. For example, starting localhost, running a Python script, toggling dark mode, and opening websites/apps are all useful commands that could be triggered with the press of a keyboard shortcut. One alternative would be bash aliases, but that doesn't come with shortcuts and instead requires pulling up Terminal, typing in your shortened alias, and then closing Terminal. The second alternative which _does_ come with shortcuts is using Automator, but that's even more tedious; to add a shortcut, you need to create a document in Automator, type your command, save it, navigate to System Preferences, and manually add a shortcut there. 

I did some idea validation on various sites and overall got some [positive responses](https://www.indiehackers.com/post/2ff5295a7c), so I definitely think that Bidbar solves the needs of, at the very least, a few. 

**Price:**
I actually was originally planning for $5, but based on [feedback](https://www.indiehackers.com/post/785fd4c687), it was suggested I raise it to $10. My only comment here is that I believe it solves a particular niche well, which makes it worth the price for some and not worth it for others. 

Obviously, Bidbar is not an app for all programmers, and in fact, some of my developer friends scoffed at the idea. However, for others, it can be a very helpful app and excel in its use cases, which is why I think it would make it a good addition. Of course, no hard feelings if it's not enough to meet the criteria 🙂

herrbischoff:
I get where you're coming from and still think the approach is a flawed one. In my view, it's a gizmo, not a tool. However, I'm but one voice regarding software on this list. I'll leave it to the others to decide.

dkhamsing:
I also do not believe this project belongs on the list. This is not very compelling to me.

herrbischoff:
Ok, let's close this then.

PR #561, opened 2020-10-06
Add Panic Nova

SalGnt:
1. [x] Project URL: https://nova.app/

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@
 - [CotEditor](https://coteditor.com) - Lightweight plain-text editor for macOS. [![Open-Source Software][OSS Icon]](https://github.com/coteditor/CotEditor/) ![Freeware][Freeware Icon]
 - [Emacs](https://www.emacswiki.org/emacs/EmacsForMacOS) - Port of Emacs to work as an macOS app. [![Open-Source Software][OSS Icon]](https://emacsformacosx.com/download/emacs-sources/) ![Freeware][Freeware Icon]
 - [MacVim](https://github.com/macvim-dev/macvim) - Vim, the text editor. [![Open-Source Software][OSS Icon]](https://github.com/macvim-dev/macvim) ![Freeware][Freeware Icon]
+- [Nova](https://nova.app/) - The beautiful, fast, flexible, native Mac code editor from Panic.
 - [Sublime Text 3](http://www.sublimetext.com/) - The sophisticated text editor.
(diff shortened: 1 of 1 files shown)


iCHAIT:
@SalGnt How is this different/better than tons of amazing editors out there? I read through the home page and I couldn't find anything in particular that is not provided by VS Code or Sublime Text already. What am I missing?

Moreover $99/year for an editor??

SalGnt:
It is really close to VS Code / Sublime Text, but native. Agree with you on the pricing, it is a big cons.

iCHAIT:
Noted. I am not sure if people really care for a native editor (at least I don't). I guess the performance and experience would be better?
Even then I think $99/year is too expensive. But that's just my opinion.

👎 from me

Up to you guys @herrbischoff @dkhamsing

dkhamsing:
Yea a bit pricey but I do see rave reviews. I guess I'll remain neutral on this.

herrbischoff:
In this case, let me be the one who supports the inclusion of Nova.

First off, it's $99 one-time and $49 a year _if you so choose_. The editor remains functional at all times, it's not a mandatory subscription, you can stop any time and resume without penalty (see: https://nova.app/buy/). I've never heard of a "subscription" that is more fair.

Apart from that, it's the only really modern Mac-centric editor that exists. VSCode for example, may be impressive but is essentially spyware (the editor itself and every single plugin constantly collect telemetry), based on Electron and not at all integrated into the system. Like always when a big corporation like Microsoft releases something so substantial for free, you need to ask yourself how this pays for them. Visual Studio (sans Code), on the other hand, is still a [$45/month subscription](https://visualstudio.microsoft.com/vs/pricing/). The others are TextMate (free, shows its age) and Sublime Text ($80, unreliable author) to some degree. 

Panic has a stellar reputation for quality and customer support as well as product support. They put their all into everything they do and are in my opinion one of the best or even _the_ best independent Mac development studio. If you want applications that are as much or more Mac-typical than Apple's own, Panic is one of very few studios who get it right. Just think of Transmit – it's practically synonymous with FTP on the Mac.

I'm writing this as a former Coda user and long-time Vim convert. Don't underestimate the usefulness of an integrated, modern editor suite on the Mac platform. Practically the only reason I'm not using it is that it's not Vim. But that goes for pretty much all editors that are not Vim. ;)

I wholeheartedly support adding Nova. For me, it represents one of the best developers on the platform creating something that is not at all guaranteed to exist any more: a very high quality Mac-typical application.

PR #589, opened 2021-08-07
Add Privacy Redirect for Safari

FIGBERT:
1. [x] Project URL: https://github.com/smmr-software/privacy-redirect-safari
2. [x] Why should this project be added: Not only is Privacy Redirect an incredibly useful extension, which I myself use every day, it also serves as an example for others seeking to make web extensions for Safari. It uses the most up-to-date frameworks from Apple, including SwiftUI, and also includes a fairly novel method for communicating between extension and native app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,7 @@
 - [OnyX](http://www.titanium.free.fr/) -  Multifunction utility to verify disks and files, run cleaning and system maintenance tasks, configure hidden options and more. ![Freeware][Freeware Icon]
 - [Paparazzi](http://derailer.org/paparazzi/) - A small utility that makes screenshots of webpages. ![Freeware][Freeware Icon]
 - [Paragon NTFS](http://www.paragon-drivers.com/ntfs-mac/) - World fastest NTFS driver.
+- [Privacy Redirect for Safari](https://github.com/smmr-software/privacy-redirect-safari) - A Safari extension that redirect sites to privacy friendly alternatives. [![Open-Source Software][OSS Icon]](https://github.com/smmr-software/privacy-redirect-safari)
 - [Radio Silence](https://radiosilenceapp.com) - Simple to use firewall and network monitor.
 - [Microsoft Remote Desktop Connection Client](https://itunes.apple.com/us/app/microsoft-remote-desktop/id715768417) - Remote Desktop Connection Client lets you connect from your Macintosh computer to a Windows-based computer.
 - [RDM](https://github.com/avibrazil/RDM) - Easily set Mac Retina display to higher unsupported resolutions. [![Open-Source Software][OSS Icon]](https://github.com/avibrazil/RDM)


iCHAIT:
Thoughts @herrbischoff @dkhamsing ?

herrbischoff:
I cannot see much value here.

It's an interesting little toy but nothing more. Also, it's a freaking browser extension. And for that it does too little. I believe there are other lists for that. Unless we decide to add a section for Safari extensions, which I propose against.

The second argument is subverted by the fact that the code is GPLv3, which means it infects every derivative building on it. I see that as a net negative, especially for code meant to learn from and be reused.

dkhamsing:
Not sure about this one

FIGBERT:
> I cannot see much value here. It's obviously a learning project from a new GitHub user, which is fine and all but not awesome.

`s/learning/side`
`s/new GitHub user/active developer who avoids GitHub when possible because Microsoft?`

> It's an interesting little toy and like so many programmers' early projects, borne out of the desire to find a problem to create a low-complexity solution for. It's a freaking browser extension. And for that it does too little. I believe there are other lists for that. Unless we decide to add a section for Safari extensions, which I propose against.

I agree with the "interesting little toy" and "[i]t's freaking browser extension" – both good assessments – but pretty much nothing else you've written, @herrbischoff. Let's start at the beginning, shall we:
1. "Borne out of the desire to find a problem to create a low-complexity solution for."
    1. It was, in fact, borne out of my own frustration. I self-host Nitter, but have to manually copy and retype Twitter URLs from Hacker News in order to use it properly. This fixes that. It's also certainly not low-complexity. I had to design a method of communicating between JavaScript extension to Swift native app with Apple APIs that are completely undocumented. I'd hazard to say this is the only codebase in the world with code that performs that function.
2. "It's a freaking browser extension. And for that it does too little."
    1. I'm sorry you don't find it useful, and think it does too little. Perhaps you should let Simon Brazell know: his [Chromium and Firefox version](https://github.com/SimonBrazell/privacy-redirect), which ours forks, has 500+ GitHub stars, 4000+ users on Chrome, 2000+ users on Firefox, and 400+ on Edge.
3. "I believe there are other lists for that."
    1. If you've got any suggestions as to where we should submit out extension, do please let us know and we'll take it into consideration.
4. "Unless we decide to add a section for Safari extensions, which I propose against.
   1. Not sure what you've got against Safari extensions. I, for one, hope the extension ecosystem thrives – I've been missing a whole host of extensions after I switched from Chrome. Extensions are useful.

> The second argument is subverted by the fact that the code is GPLv3, which means it infects every derivative building on it. I see that as a net negative, especially for code meant to learn from and be reused.

Yeah, I like to use GPL – this is a separate debate that has been had thousands of times. I will say that the GPL does not prevent people from learning from my code: it does prevent them from releasing a closed source version, which is a feature I'm a fan of.

More generally: this is your repository and your project. You can do what you like with it, including not adding Privacy Redirect for Safari to the list. I don't really care: but I must say, I found @herrbischoff's remarks rude and uncalled for. It's one thing to, as @dkhamsing did, not support its addition to the list. It's another to baselessly insult the project and its makers. An apology is in order.

herrbischoff:
Please don't create drama where there is none.

PR #605, opened 2021-11-28
add Strongbox password manager

beerisgood:
1. [x] Project URL: https://strongboxsafe.com/
2. [x] Why should this project be added: best offline password manager. awesome


herrbischoff:
This may well be a fine piece of software, there is however no way to test it, as there is no test version and it is commercial.

I see two ways forward: either we get several endorsements from different users that this is indeed great software or you convince the application's author to give the team licenses (temporary ones are fine) to find out for ourselves.

stackptr:
> there is however no way to test it, as there is no test version and it is commercial.

I downloaded the macOS version from the App Store link on the home page and it greeted me with a message that I had 3 months to use the "Pro" features prior to [downgrading](https://strongboxsafe.com/comparison/). I have no intention of testing or endorsing, but there does seem to be a method for others to do so.

At the very least, it doesn't appear to use Electron.

iCHAIT:
LGTM

PR #607, opened 2021-12-31
Add Speediness

sindresorhus:
1. [x] Project URL: https://sindresorhus.com/speediness
2. [x] Why should this project be added: It's a common need to want to know your internet speed, and this app does it in a very simple way. It also comes with powerful features like Shortcuts support and history. Most other tools for this are biased (for example, Speedtest by Ookla). Speediness returns more realistic result by running the tests in parallel. [More here](https://sindresorhus.com/speediness#faq)

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@
 - [SlowQuitApps](https://github.com/dteoh/SlowQuitApps) - Prevent accidental Cmd-Q. [![Open-Source Software][OSS Icon]](https://github.com/dteoh/SlowQuitApps) ![Freeware][Freeware Icon]
 - [SmartCapsLock](https://kishanbagaria.com/smartcapslock/) - Makes the Caps Lock key smarter, so that when the key accidentally gets activated and you START YELLING even though you don't want to, you can just select the yelling-text and press the key again to instantly fix its case instead of typing everything all over again.
 - [Soulver](http://www.acqualia.com/soulver/) - Beautiful expressive calculator.
-- [Speediness](https://apps.apple.com/app/id1596706466) - Check your internet speed. ![Freeware][Freeware Icon]
+- [Speediness](https://apps.apple.com/app/id1596706466) - Check your internet speed (requires Monterey). ![Freeware][Freeware Icon]
 - [SSH Tunnel](https://codinn.com/products/ssh-tunnel/) - Manage your SSH tunnels, tightly integrated with macOS Keychain, secure and intuitive.
 - [TeamViewer](https://www.teamviewer.com/en/) - Remotely control another computer.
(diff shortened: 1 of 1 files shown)


herrbischoff:
> The app makes use of the built-in `networkQuality` command-line tool under the hood, which is provided by Apple.

Why do we need a GUI wrapper for that? Also, `networkQuality` was introduced and is exclusive to Monterey. Something to keep in mind here.

sindresorhus:
The fact that it uses a command-line tool underneath is just an implementation detail. The app is for any user, not just developers who are comfortable with the command-line. The app also includes more functionality than `networkQuality`, so it's useful for developer too; It includes ping, history and export.

herrbischoff:
At the very least, the description should reflect that it requires Monterey. Not everyone wants to use the latest OS or even *can* use it. In my view, this is not "just an implementation detail", it's the entire deal.

To illustrate this, please allow a little side note, as this affects me directly. My Mid 2014 MBP is not supported by Monterey, despite still being in perfect working order and still being fast enough for a main work machine. Therefore I cannot even check out your application as it requires a command-line application that — for seemingly no reason at all — is only available on the latest OS. That's Apple and it is what it is.

If this application somehow embedded this functionality for older OS versions as well, I'd consider that "awesome". As is, to me, it's just a fancy wrapper. Therefore I'd still question its utility but leave it up to the others to decide. This is not meant as a confrontation, just some food for thought what should pass as "awesome" and what maybe shouldn't. With you being the initiator of the Awesome lists and all.

sindresorhus:
> At the very least, the description should reflect that it requires Monterey. 

👍 Done

sindresorhus:
> If this application somehow embedded this functionality for older OS versions as well, I'd consider that "awesome".

Whether or not an app targets the latest OS is a choice the app is allowed to make, and does not make it more or less awesome. There are many valid technical reasons to target the latest OS. There might be new APIs the app needs. There are also maintenance related reasons, an app could have backwards compatibility, but it requires a lot of effort, which does not always make sense, for business or personal reasons.

Personally, I always target the latest current OS version when I publish a new app. This lets me take advantage of new APIs and also means less OS bugs, which in turn makes it much more enjoyable and faster for me to develop apps. I mainly make apps for my own personal enjoyment.

What you should actually look for in an "awesome" app:
- Performance
- Ease of use
- Whether it looks good
- Whether it follows the macOS Human Interface Guidelines
- App Store rating
- Whether it's actively maintained

sindresorhus:
You should not judge an app based on whether it's useful to you. There are lots of apps you don't need that are awesome to a large percentage of other users.

PR #614, opened 2022-03-29
Added a new category called 'Remote Tools' (as lot of us are wfh nowadays)

rampatra:
1. [x] Project URL: I have added the URLs for the respective apps.
2. [x] Why should this project be added: These are great apps used by many with awesome developers behind them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@
     - [Graphics](#graphics)
     - [News Readers](#news-readers)
     - [Productivity](#productivity)
+    - [Remote Tools](#remote-tools)
     - [Sharing Files](#sharing-files)
     - [Terminal](#terminal)
     - [Utilities](#utilities)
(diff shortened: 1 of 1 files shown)


herrbischoff:
I don't see the need for an arbitrary section like this. The examples moved there concern presentations and video conferencing. Both areas existed well before widespread WFH and independent of this particular use case.

dkhamsing:
@rampatra maybe we can consider adding Lungo and Meeter ? Could you create one pull request per suggestion

Thx!

iCHAIT:
Please open a separate PR for each recommendation. Also, I think it would be better to add these tools in either Productivity or Utility.

herrbischoff:
Don't know about you but I really have to say I'm growing tired of all the malformed drive-by pull requests.

PR #616, opened 2022-05-20
Add Wavebox

Thomas101:
Add the Wavebox app to the productivity section

1. [x] Project URL: https://wavebox.io/download
2. [x] Why should this project be added: Wavebox is a Chromium browser (not Electron) that was first launched for macOS in 2016. It allows you to add and organize your most recently used sites into the sidebar for fast access


herrbischoff:
According to the website this is "freemium" software, asking for a recurrent subscription payment to use it as intended, which is €12 + VAT/month. I haven't tested it but then again, I see no point in the web-appification of everything in the first place. In my book, if you feel the need for a separate piece of software/browser just to keep all the web apps in check that you use on a daily basis, something is deeply wrong.

Anyway, I feel that it is way too expensive for what it does and requires yet another account at yet another service with yet another subscription. Existing browser tab pinning provides me with most of that for free, no third-party account required.

@Thomas101, as you appear to be the project's main developer, I wish you best of luck with the approach, even if I believe it to be a fundamentally flawed one. Different people will undoubtedly see this differently.

So unfortunately that's a 👎 from me. What do you think, @iCHAIT @dkhamsing?

dkhamsing:
Would be a no as well.. not compelling based on the description provided above.

PR #626, opened 2023-01-02
Add DevUtils app

trungdq88:
1. [x] Project URL: https://devutils.com
2. [x] Why should this project be added: I think it's an awesome native macOS app and useful for many developers. It's being used by 10,000+ developers and teams around the world. It would be great if we can include it in this repo, thanks! (Disclosure: I'm the author of the app).


herrbischoff:
While it looks potentially useful, two comments:

- "Because it's awesome" is not a valid reason for inclusion at all, especially if you are the author, very especially if it's a commercial application. When you're this lazy, I will be too.
- The dark pattern of download buttons everywhere on the website but pricing tucked away in a menu item feels wrong. I'd also argue that $29 for a bunch of parsers and converters is rather steep.

I'm not the intended target consumer because I think there's an inverse correlation between emoji frequency and utility. Therefore I leave this up to the others.

trungdq88:
Whoops, sorry about the short description. I wasn't sure what was expected in that question so I put a generic one. I'll update.

PR #634, opened 2023-05-02
Remove not existing app

beardcoder:
Paw website is down


herrbischoff:
This is a nonsensical pull request as well. Instead of deleting an entry outright, please put in even the least bit of effort to justify the decision.

Google immediately surfaces: https://paw.cloud/

PR #636, opened 2023-06-09
Change title Fantastical 2 -> Fantastical

SoneeJohn:
Just a simple title change.


herrbischoff:
Thanks. Good to see one of the devs care about the details. This is in line with my experience with Fantastical.

PR #640, opened 2023-06-28
Add pgMagic

hill:
1. [x] Project URL: https://pgmagic.app
2. [x] Why should this project be added: pgMagic is a small (14mb) mac postgres client that lets you do anything you would expect from a regular db gui, but also lets you query postgres in natural language. Its super helpful for data and business analysts that have some understanding of SQL, but don't want to craft custom queries for every minor question asked of them. Its also very helpful for people new to SQL to get a basic query up for them to edit and alter. Being a mac app, no db credentials are sent to a third party server and are kept locally on your machine. Personally I have found it extremely valuable in quickly gaining insights into user actions that are somewhere in my db but not easily accessible and would require a relatively hefty query to obtain (saving me 10 to 15 minutes).
4. [x] End all descriptions with a full stop/period
5. [x] Entry is ordered alphabetically
6. [x] Appropriate icon(s) added if applicable (OSS, freeware)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@
 - [Medis](https://getmedis.com) - A modern GUI for Redis.
 - [Pasteboard Viewer](https://apps.apple.com/app/id1499215709) - Inspect the system pasteboards. [![Open-Source Software][OSS Icon]](https://github.com/sindresorhus/Pasteboard-Viewer) ![Freeware][Freeware Icon]
 - [Paw](https://luckymarmot.com/paw) - The ultimate REST client.
-- [PgMagic](https://pgmagic.app/?ref=awesomemacos) - A Postgres client that lets you talk to your db in SQL or natural language.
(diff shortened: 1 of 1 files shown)


hill:
@stackptr cheers - thanks 💙

herrbischoff:
The referrer query string needs to be removed. This bad form. Tracking of any kind is unacceptable to me. The fact that no other entry has this should have been a hint.

It also makes me think what kind of telemetry this app collects.

hill:
Hi @herrbischoff - no worries, it's removed. You're right should have taken a hint from the other listings. I usually use the referrer string to provide a discount to customers coming from these aggregation lists. 

> It also makes me wonder what kind of telemetry this app collects.

None. Its a local app specifically so that users didn't have to send database credentials to some third party server. It does log **locally**, however these are never automatically collected (the user can choose to send this log file to me if they wish to help resolve a bug)

herrbischoff:
Good to know, thanks.

PR #646, opened 2023-10-15
Add Lasso - window manager

iSapozhnik:
1. [x] Project URL: https://thelasso.app/
2. [x] Why should this project be added: The app implements an easy-to-use grid-based approach for window management. It's maintained (new updates coming almost every month), native, and supports macOS 12 - macOS 14 Sonoma.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,7 @@
 - [Divvy Window Manager](http://mizage.com/divvy/) - Window management for tiling your windows.
 - [Hammerspoon](http://www.hammerspoon.org/) - Extremely powerful scripting engine for macOS. [![Open-Source Software][OSS Icon]](https://github.com/Hammerspoon/hammerspoon) ![Freeware][Freeware Icon]
 - [Hummingbird](https://hummingbirdapp.site/) - Easily move and resize windows without mouse clicks, from anywhere within a window.
+- [Lasso](https://thelasso.app/) - Grid-based intuitive window manager for macOS.
 - [Moom](https://manytricks.com/moom/) - Move and zoom windows, super light weight and customizable.
 - [Phoenix](https://github.com/kasper/phoenix) - A lightweight window and app manager scriptable with JavaScript. [![Open-Source Software][OSS Icon]](https://github.com/Hammerspoon/hammerspoon) ![Freeware][Freeware Icon]
 - [Rectangle](https://rectangleapp.com/) - Easily organize windows without using a mouse. [![Open-Source Software][OSS Icon]](https://github.com/rxhanson/Rectangle) ![Freeware][Freeware Icon]


iCHAIT:
NOt sure about this. Will leave it up to @herrbischoff and @dkhamsing  to decide

iSapozhnik:
@iCHAIT What makes you doubt?

iCHAIT:
Hi @iSapozhnik 

I use Rectangle and love it, it is already on the list as well. It is free and open source and works amazingly well. I am not an advanced user when it comes to Windows Manager so maybe some other maintainers would have better judgment on this. Also, the pricing of the tool makes me hesitant to give it a shot when I already have something that works well (for free).

herrbischoff:
Maybe I'm old but I increasingly step away from small, fancy applications and stay with simple tools that work. Like Rectangle. It really does everything I need. 

I've also been burned one time too many by productivity tools from single developers that I built my workflow around that didn't survive macOS updates because they lost interest in fixing it.

My vote is no.

iSapozhnik:
@iCHAIT @herrbischoff 
Hey all!
I probably made a mistake submitting this app to the list in the first place because apparently it's not about awesome apps. The app should pass the review by people who might not be even a target audience. I will close the PR but would like to address some of your concerns.

> Also, the pricing of the tool makes me hesitant to give it a shot
Your guidelines say "Commercial applications should offer a free trial for maintainers to validate inclusion to the list." and the app does offer a free 7-day trial period. On the main page of the site, there is just one button - Download, and everyone who wants to give it a shot could have done it in one click. Since @iCHAIT in his message expressed a couple of times his appreciation of the fact that Rectangle is free, to me it sounds like commercial apps are not welcomed here. Otherwise, why do you emphasize it - almost every open-source app is free.  

@herrbischoff what you are saying does make a lot of sense. However, it contradicts the idea of the repo (at least if my understanding of "awesome applications and shiny things" is correct). Just because, as you say, you are old and used to one tool you are vetoing the app to be in the list that is supposed to be a place where everyone (community members) could find a suitable app for their needs. I'm also not the youngest person on the planet but I keep using such repositories to explore new apps to fulfill my needs that tend to change over time. I believe there are thousands of people like me. You also mentioned that you try to step away from small and fancy applications from single developers because eventually, they might lose interest in developing them. This applies to any application no matter if it is from a single developer or a company. There are a lot of such examples but I will use the one that is already in this repo - Divvy. This app is made by a company and has not been updated for years, does not support the latest OS and Apple processors, and has a lot of bugs. Of lot of Lasso users are actually grateful that it exists because it is like the next step after Divvy.

And my last point. In your messages, you both say about the rectangle app (which is a freaking amazing piece of software - heads down to @rxhanson and all the contributors) and that it fits YOUR needs. Other people might have a completely different set of requirements. And that is again where I personally see a dissonance between "a list of awesome applications" and "applications loved by @iCHAIT and @herrbischoff".

PR #658, opened 2024-02-18
Add ChromeBuddy

AndrejSoroj:
1. [x] Project URL: https://github.com/AndreyGr/ChromeBuddy
2. [x] Why should this project be added: ChromeBuddy is tiny app that is set as default browser and allows to open external links in Chrome's top window not matter if it is incognito or common window(the default Chrome behaviour does not allow to open external links in incognito window). Overall this application addressing the following fix in Chromium engine: https://issues.chromium.org/issues/40536115 that is changed original behaviour which this app is intended to replicate. The motivation of adding this app is fare amount of users as me that complaining on the change in Chrome behaviour and looking for solution to revert original behaviour.


stackptr:
Seems like a learning project with very few users. Given the niche use-case, I'm concerned that adding it will just create noise.

PR #736, opened 2026-03-25
Add "AI" related wording

herrbischoff:
Due to the increasing amount of "AI" adjacent pull requests, I've added clarifying language regarding the handling of such submissions. I advocate strongly for excluding them, just like Electron apps. One would think the reasons for rejecting generative slop are obvious enough at this point.

Additionally, due to an increase of pull requests that just skip the template entirely, I've added language that makes the consequences of lazy pull requests clear.

@iCHAIT, please review and merge if appropriate.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,14 +3,16 @@
 <!-- Please fill out the following: -->
 
 0. [ ] I have read the [Contribution Guidelines](https://github.com/iCHAIT/awesome-macOS/blob/master/.github/contributing.md)
-1. [ ] Project URL:
-2. [ ] Why should this project be added:
-3. [ ] End all descriptions with a full stop/period
-4. [ ] Entry is ordered alphabetically
-5. [ ] Appropriate icon(s) added if applicable (OSS, freeware)
+1. [ ] I confirm that I have read and understood the sections about "AI"-adjacent software
+2. [ ] Project URL:
+3. [ ] Why should this project be added:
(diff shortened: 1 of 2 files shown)


iCHAIT:
Hi @herrbischoff 

Thanks for putting this together. Totally agree we need to address the low-effort submissions, especially the ones missing the template altogether.

On the AI exclusion, I'm on board with keeping out low-quality "AI wrapper" apps, but I'm a bit hesitant about a blanket "No AI anything" rule. I wouldn't want to exclude genuinely good apps just because they mention AI somewhere.

What do you think about narrowing the scope a bit? Something like:
`No apps whose primary purpose is wrapping a generative AI API` or
`No AI-first apps that are essentially chatbot interfaces or prompt wrappers`

Let me know what you think?

herrbischoff:
Thanks for your response. I feel that being vague or permissive will lead to discussions with annoying boosters.

ML is distinct from "AI" in my view, especially specialised local models trained on hand-picked data. Those are great. But anything LLM generated like ChatGPT or Claude is an instant disqualification for me, even if used for summaries. If this is optional functionality within a larger scope it's fine but anything explicitly catering to generative AI should be excluded.

How about this?

> No AI prompt wrappers or software whose main purpose is interfacing with generative tools like LLMs, chatbots, code generation APIs and similar, including agents. Leveraging local, specialised ML models as part of a larger process is fine. Think image generation via prompt vs. manual image editing suite featuring a smart repair brush.

iCHAIT:
Sounds good to me @herrbischoff 👍🏻

iCHAIT (approved):

PR #740, opened 2026-03-27
Add TouchBridge — free Touch ID alternative for Macs without biometric sensor

HMAKT99:
Title:
  Add TouchBridge — free Touch ID alternative for Macs without biometric sensor

 Description:
  

  

  1. [x] Project URL: https://github.com/HMAKT99/UnTouchID
  2. [x] Why should this project be added: TouchBridge solves a real gap in macOS — Macs without Touch ID (Mac Mini, Mac Studio, Mac Pro, upcoming MacBook Neo) have no biometric auth option except buying a
  $199 Magic Keyboard. TouchBridge uses your phone's fingerprint/Face ID instead. Free, open source, 91 tests, works with iPhone, Android, Apple Watch, Wear OS, or any browser. No cloud dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@
 
 
 ### Productivity
-
+- [TouchBridge](https://github.com/HMAKT99/UnTouchID) - Use your phone's fingerprint to authenticate on any Mac. Free alternative to $199 Touch ID keyboard. [![Open-Source Software][OSS Icon]](https://github.com/HMAKT99/UnTouchID) ![Freeware][Freeware Icon]
 - [Alfred](https://www.alfredapp.com/) - Boosts your efficiency and productivity.
 - [BetterTouchTool](https://folivora.ai) - Configure gestures for mouse and actions for keyboard shortcuts.
 - [ClipMenu](http://www.clipmenu.com/) - ClipBoard History Manager. [![Open-Source Software][OSS Icon]](https://github.com/naotaka/ClipMenu) ![Freeware][Freeware Icon]


stackptr:
1 commit made 2 hours ago. 👎🏼

HMAKT99:
The repo was recently restructured — commit history was consolidated for a clean public launch. The project has 91 tests, a working PAM module (verified on real hardware), iOS/Android/watchOS/WearOS companion apps, and a .pkg installer. Happy to answer any questions about the codebase.

herrbischoff:
While certainly an interesting project, it becomes instantly unattractive due to the requirement of having to build everything from scratch – including the iOS companion app, as opposed to just installing and running a piece of software.

I get why this is but it also severely limits its use to developers enrolled in the $99/year Apple developer program to be able to do so as intended.

Also, moving an unverified and unauthorised piece of software in the middle of an admin auth process does not sound like something the vast majority of users should ever consider doing, much less normalise.

None of us is qualified enough to review a code base for security best practices. While the functionality is certainly useful to have, it's best reserved for the tinkerers and hackers. This list is not that.

If this was a pair of App Store applications with a verified track record, I'd sign off on them immediately and probably use them myself. As things stand, the danger and inconvenience is just too great.

HMAKT99:
Thanks for the feedback @herrbischoff. Just to clarify — users don't need Xcode or a developer account:

1. **One-click .pkg installer** — download, double-click, done: https://github.com/HMAKT99/UnTouchID/releases/tag/v1.0.0
2. **Web companion** — works with any phone's browser, no app install: `touchbridged serve --web`
3. **Simulator mode** — works with no phone at all for testing

Xcode is only needed if someone wants to build the native iOS app from source. The web companion gives the same functionality without it.

On security — PAM uses `auth sufficient`, so password always works as fallback. Code signing and notarization for the .pkg is on the roadmap.

Happy to address any other concerns!

herrbischoff:
Thanks for your response and additional info.

> 2. **Web companion** — works with any phone's browser, no app install: `touchbridged serve --web`
> […]
> Xcode is only needed if someone wants to build the native iOS app from source. The web companion gives the same functionality without it.

Wait, so admin privilege authentication happens via an open HTTP connection on a Wi-Fi connection in a web browser? Not only does that mean both iPhone and Mac need to be connected to the same network. Especially in public settings (in the README you mention coffee shops), this sounds like cavalier security practice to me.

To the best of my knowledge, Apple does something like this via an encrypted Bluetooth connection with added proximity measurement.

HMAKT99:
Good catch @herrbischoff — you're right to flag that.

The web companion is a **convenience mode for quick testing**, not the recommended production setup. The primary auth mode is exactly what you described:

**BLE (Bluetooth Low Energy)** — encrypted, local, no Wi-Fi:
- ECDH ephemeral session keys per connection
- AES-256-GCM encrypted channel
- RSSI proximity gate (configurable threshold, default -75 dBm)
- No network/Wi-Fi involved at all

The web companion (`--web`) is meant for "try it in 60 seconds" demos and cases where someone doesn't want to install a companion app. It uses one-time tokens (32 bytes, 60s expiry) and only works on the local network — but you're right that it's not suitable for public Wi-Fi or security-sensitive environments.

I should make this distinction clearer in the README — BLE is the secure path, web companion is the convenience path. Good feedback, will update.

herrbischoff:
Good to know. This however brings us back to my original point: the need to compile and install the iOS app. Unless and until the companion app is available for download through official channels (App Store), this amounts to a tech demo, as opposed to stable software ready to use.

I'll be closing this for now. Feel free to submit again once it has matured sufficiently and doesn't require compiling anything by the user.

PR #750, opened 2026-03-31
Fix broken links and missing ToC entry

arpitjain099:
## Summary
- Fix Pusher OSS icon linking to the wrong repository (was pointing to PostgresApp/PostgresApp instead of noodlewerk/NWPusher)
- Update IRC channel links from defunct Freenode to Libera.Chat (Freenode shut down in 2021)
- Add missing macOS 10.15 Catalina entry to Table of Contents (section exists in body but was missing from ToC)

## Details
Found via automated link and structure analysis. No content changes beyond fixing references.


herrbischoff:
Have the common courtesy not to use a LLM. If you don't think it's worth your time to write it yourself, it's not worth my time to read it.

PR #763, opened 2026-04-04
Add DockWright - Native macOS AI assistant

AdelElo13:
DockWright is a native macOS AI assistant built in Swift with zero dependencies. Features: screen awareness, voice control, 30 system tools, 9 LLM providers. https://dockwright.com


herrbischoff:
Every additional submission of AI slop despite clearly stating in multiple places that we don't want it here strengthens the argument for the growing mental deterioration of its users.

PR #764, opened 2026-04-04
Add Typer - free local AI chat for macOS

duarte:
Typer is a free, local AI chat app for macOS (Apple Silicon) that runs entirely on-device — no account, no subscription, no data sent to any server. It works fully offline and includes an on-device web search feature.

It fits the Chat Clients section as a native macOS AI chat client, and qualifies for the Freeware badge since it's completely free with no paid tier.


herrbischoff:
Every additional submission of AI slop – despite clearly stating in multiple places that we don't want it here – strengthens the argument for the growing mental deterioration of its users.

PR #765, opened 2026-04-05
Add Dimly

Punshnut:
0. [x] I have read the Contribution Guidelines
1. [x] I confirm that I have read and understood the sections about "AI"-adjacent software
2. [x] Project URL: https://feuerbacher.me/projects/dimly
3. [x] Why should this project be added: It allows controlling brightness across mixed internal and external monitor setups on macOS.
4. [x] End all descriptions with a full stop/period
5. [x] Entry is ordered alphabetically
6. [x] Appropriate icon(s) added if applicable (OSS, freeware)


herrbischoff:
I just tried it and am unconvinced. The idea is a good one, the execution rather lackluster.

For one, I don't understand the UI choices. Maybe on setups with more than three displays, this plays to its strengths. Then again, the software didn't recognize my Dell monitors as DCC capable while Monitor Control (which I usually use) absolutely does. It's should be a tool that does one thing and does it well. What it currently does is one thing and does it… alright. When it has matured a bit, we can certainly revisit it.

Punshnut:
Thanks for taking a look, really appreciate the feedback.

The DDC issue is interesting - I’ll dig into Dell compatibility, that should definitely work. I also get your point on the UI, there’s probably room to simplify things a bit.
Happy to revisit this later once it’s a bit more mature.

PR #787, opened 2026-04-21
Add 不拖 (Butuo) — free minimalist to-do app for iPhone & iPad

Oliverwqcwrw:
## What is 不拖 (Butuo)?

A minimalist to-do app for iPhone and iPad that fits perfectly in the Productivity section alongside OmniFocus and Taskade.

### Why it belongs here
- 🍎 Native iOS app (iPhone + iPad)
- **Completely free** — no subscription, no in-app purchases
- **No account, no server, no sync** — all data stored locally on device only
- Zero-friction UI: open the app, write it down, cross it off
- No ads, no tracking

**App Store:** https://apps.apple.com/app/id6761042128


iCHAIT:
Not a contender for macOS. 

You may want to consider https://github.com/vsouza/awesome-ios

Thanks

PR #831, opened 2026-05-26
Add ModelStatus to Utilities

lucasmullikin:
Adds ModelStatus to the Applications → Utilities section.

ModelStatus is a free, open-source (MIT) macOS menu bar app for monitoring multiple local LLM servers — Ollama, LM Studio, vLLM, MLX, llama.cpp, and any OpenAI-compatible API. Multi-instance, auto-provider-detect, LAN + Tailscale peer discovery. Pure Swift/AppKit, no dependencies, no telemetry.

- Source: https://github.com/lucasmullikin/ModelStatus
- Install: `brew tap lucasmullikin/tap && brew install --cask modelstatus`
- macOS 13+, MIT license


herrbischoff:
Every additional submission of AI slop – despite clearly stating in multiple places that we don't want it here – strengthens the argument for the growing mental deterioration of its users.

PR #904, opened 2026-07-05
Remove Juicy due to excessive telemetry collection

herrbischoff:
The application aggressively attempts to collect telemetry data via TelemetryDeck.

My Pihole registers upwards of 5k connection attempts per 24h window.

Screenshot 2026-07-05 at 07 57 41

This only started showing up after I had installed the app. I had asked the author @rylax via email for a comment, to no avail. Looking into it myself, I verified that Juicy is the only application on my computer using TelemetryDeck.

Screenshot 2026-07-05 at 07 56 22

While Juicy's functionality is nice, forcing non-optional telemetry collection on users, especially this aggressive, disqualifies any application for me. Therefore, I propose removing Juicy until and unless this is rectified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,7 +296,6 @@
 - [Itsytv](https://itsytv.app) - Control your Apple TV from the menu bar with remote, now playing widget, and app launcher. [![Open-Source Software][OSS Icon]](https://github.com/nickustinov/itsytv-macos) ![Freeware][Freeware Icon]
 - [iStat Menus](https://bjango.com/mac/istatmenus/) - An advanced system monitor for your menubar.
 - [Jiffy](https://sindresorhus.com/jiffy) - Discover and share the best GIFs on GIPHY. ![Freeware][Freeware Icon]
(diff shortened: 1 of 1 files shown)


rylax:
Hey Marcel, just followed up via email as well.

TelemetryDeck is built by two privacy-focused developers in Germany and all data is fully anonymous. All it really does is check which version of Juicy is installed and what macOS version it's running on, so I can make better decisions around features and compatibility. Nothing beyond that.

That said, I totally understand where you're coming from. The next update will include an opt-in during onboarding and a toggle in Settings to turn TD off permanently.

Appreciate you flagging this.

iCHAIT:
Hi @rylax 

Any timeline for this?

rylax:
@iCHAIT Hi there! Update is already live since 1.5.3 :) There's a toggle with very declarative info about the full anon data share option during onboarding as well as in the general settings a toggle to opt in and out whenever desired.

iCHAIT:
Thanks for the update @rylax .

@herrbischoff, I am closing this now. Thanks.

PR #975, opened 2026-08-06
Add Eqlume

gokturkgocen:
Adding [Eqlume](https://github.com/gokturkgocen/Eqlume) — a free, open-source (MIT) system-wide equalizer for macOS, now also on the [Mac App Store](https://apps.apple.com/app/id6793070613).

What sets it apart from the other audio entries: rather than making you manage presets, it detects the genre of each track as it plays (MusicBrainz → iTunes Search → an on-device CoreML classifier) and applies a matching curve automatically. It uses a Core Audio process tap, so there's no virtual audio driver or kernel extension.

Placed alphabetically between BackgroundMusic and Fader, with the OSS + Freeware badges used by neighbouring entries. One line, no other changes.


gokturkgocen:
Good catch — changed to "EQ curve", which better reflects that nothing needs to be managed by hand.

